PR #10, opened 2023-02-01
Add support for LCD display and remove OLED

hnez:
This PR is based on #1 and should be rebased on top of it if there are any changes there.

This PR is marked as WIP for a few reasons:

- [ ] Bundles built with this branch only work on LXA TACs with an LCD instead of an OLED screen of which there are currently: zero
- [x] The generic MIPI DBI driver uses a firmware file to initialize the display. The file has to be available on probe so the driver needs to be built as module and is currently manually `modprobe`d. Check if there is a nicer way to do this.
- [ ] The `tacd` commit this PR is based on is not yet in main. The respective PR linux-automation/tacd#1 should be merged first
- [ ] The panel-mipi-dbi recipe should be split into two recpies, one for the tool and one that uses the tool to build the firmware.


jluebbe (changes requested):

hnez:
This PR is superseded by #33.
Open tasks have been transferred.
Progress will now be tracked there.

PR #14, opened 2023-02-16
meta-lxatac-software: labgrid: allow overriding USB ports in userconfig

hnez:
This should make it easier to attach USB hubs to the TAC and getting the matches right.

I'm not too experienced with jinja2 so I've had some trouble working around the scoping rules (`set`ing a variable in a `include`d file does not have an effect on the file that did the `include`ing, but modifing the variable in another way does).
And preventing the output from my `usb_ports.pop` and `.extend` tricks from being rendered (that's what the `and '' and `or ''` are for).
Maybe you have some better idea.

@a3f could this be of any use to you?


SmithChart (approved):

jluebbe (changes requested):
Looks good otherwise.

jluebbe (approved):

SmithChart (approved):

PR #24, opened 2023-02-24
meta-lxatac-bsp: base-files: add fstab with explicit mount for /srv

hnez:
We have previously relied on `systemd-gpt-auto-generator` to automatically generate a `srv.mount` unit based on the partition type of the data partiton. The data partition is created on first boot by `systemd-repart`.

This results in `/srv` not being mounted on first boot as `systemd-generators` run very early in the boot process, long before `systemd-repart` had a chance to create the partition.

Add a `fstab` (mostly the default `fstab` from poky) that includes an explicit `/srv` mount.


hnez:
PR #42 is a cleaner attempt to solve the same issue.
I will close this PR.

PR #31, opened 2023-03-03
labgrid-exporter: export FTDIs with multiple separate interfaces

a3f:
Some FTDIs pop up as multiple USB interfaces on the host system. Let's export there as well. This could've been done in the userconfig.yaml, but it's more convenient to just do this here centrally.

We'll just export the main interface and three more and assume 4 FTDIs is enough for everybody.


SmithChart:
This is exactly the same setup we are already using in production on our lab system. LGTM

PR #33, opened 2023-03-08
Support LXA TAC Gen 2 devices (BSP)

hnez:
Hardware generations
-----------------------------

There is a new batch of LXA TAC devices with a slightly different set of hardware features (**Gen 2**).
Here is an overview of the current hardware generations:

- **Gen 1 Stock**
  - 128px x 64px OLED screen
  - No LED indicators on the top side of the device.
- **Gen 1 LCD**
  - Like **Gen 1 Stock** but retrofitted with a 240px x 240px LCD screen
  - All of the **Gen 1 Stock** devices should be modified to this state
- **Gen 2**
  - 240px x 240px LCD screen
  - LED indicators on the top side of the device

With this PR applied only hardware generations **Gen 1 LCD** and **Gen 2** are supported. Support for **Gen 1 Stock** is dropped.

Related pull requests
----------------------------

- `meta-lxatac`: This PR continues the work started in #10 to support **Gen 1 LCD** and it will thus be closed. There is however an open tasks there that have not yet been addressed:
  - [x] The panel-mipi-dbi recipe should be split into two recpies, one for the tool and one that uses the tool to build the firmware.
- `tacd`: `tacd`s with the PR linux-automation/tacd#13 applied depend on the devicetree changes made in this PR to run. Merging of the PRs should be done in tandem. In addition to that:
  - [x] Update the `tacd` recipe to use the upstream repo and `main` branch with the PR applied.
  - [x] Update the `tacd-webinterface` recipe to use the upstream repo and `main` branch with the PR applied.

Other open tasks
-----------------------

- [x] Decide how the RGB status LED on Gen 2 devices should behave during boot. Do we e.g. want:
  1) Set a "in progress" color in barebox and change to an "error" color if something fails.
  2) Set another "in progress" color in linux or start pulsing in the previous color to indicate the bootloader -> linux change.
  3) Set a "sucess" color in linux when e.g. `boot-complete.target` is reached. Or should this be done in the `tacd`?
- [x] Clean up …
- [ ] … and review the barebox device detection and devicetree selection
- [x] Clean up …
- [x] … and review the linux devicetree / defconfig changes.
- [x] This PR also upgrades the linux kernel to `6.3` make sure that is released and the recipe is updated to the released version. Or backport the changes to `6.2`.
- [x] The `tacd.service` currently uses fixed `gpioset` commands to turn off the `DUT` on failure. This does not work with the Gen 2 powerboard which does not have a GPIO expander. Find a different solution.
- [x] When interacting with a Gen 1 device over the debug port it seemed that barebox as become a bit hard of hearing, e.g. missed a lot of characters. Investigate/fix that.
  20230418 lgo: @SmithChart has found out that this is caused by a floating GPIO line for the input button on the powerboard if no power board is connected. The solution is to enable internal pullups for the buttons.


meta-lxatac-bsp/recipes-kernel/linux/files/defconfig
```diff
@@ -1704,6 +1702,7 @@ CONFIG_NET_VRF=y
 CONFIG_NET_DSA_MICROCHIP_KSZ_COMMON=y
 # CONFIG_NET_DSA_MICROCHIP_KSZ9477_I2C is not set
 CONFIG_NET_DSA_MICROCHIP_KSZ_SPI=y
+CONFIG_NET_DSA_MICROCHIP_KSZ_PTP=y
```
olerem: I would not recommend to enable KSZ_PTP. This change will cost about 50Mbits/s
Here are iperf3 results with KSZ_PTP enabled:
[ ID] Interval           Transfer     Bitrate         Retr                                                                                               
[  5]   0.00-100.00 sec  7.09 GBytes   609 Mbits/sec  140             sender   
[  5]   0.00-100.04 sec  7.09 GBytes   609 Mbits/sec                  receiver
Here are iperf3 results with KSZ_PTP disabled:
[ ID] Interval           Transfer     Bitrate         Retr                                                                                               
[  5]   0.00-100.00 sec  7.68 GBytes   660 Mbits/sec  186             sender     
[  5]   0.00-100.04 sec  7.68 GBytes   659 Mbits/sec                  receiver

meta-lxatac-bsp/recipes-kernel/linux/files/defconfig
```diff
@@ -1704,6 +1701,7 @@ CONFIG_NET_VRF=y
 CONFIG_NET_DSA_MICROCHIP_KSZ_COMMON=y
 # CONFIG_NET_DSA_MICROCHIP_KSZ9477_I2C is not set
 CONFIG_NET_DSA_MICROCHIP_KSZ_SPI=y
+# CONFIG_NET_DSA_MICROCHIP_KSZ_PTP is not set
```
hnez, line 1704: Thanks to the tip from @olerem I've now disabled this feature.

meta-lxatac-bsp/recipes-kernel/linux/files/defconfig
hnez: Thanks for the tip and measurements. I've disabled the feature in the newest version of this PR.

jluebbe:
I've looked at the Linux defconfig changes and found no issues.

Emantor (approved):

PR #34, opened 2023-03-22
meta-lxatac-software: android-tools: backport 10.0.0 from meta-oe

hnez:
The `fastboot` binary we currently ship only works when manually supplied the USB vendor id of the device we want to flash.
This is non-intuitive and newer `fastboot` just finds the right device automagically.

Backport a newer `android-tools` [from meta-oe master](https://github.com/openembedded/meta-openembedded/tree/b71b1ee/meta-oe/dynamic-layers/selinux/recipes-devtool/android-tools) and add `meta-selinux` which provides `libselinux`.

I'm not a fan of adding `meta-selinux` just for `libselinux`, so if there are any other ideas I am open for it.

The addition of the meta layer also means that bblayer templates in custom `meta-lxatac` derived BSPs have to be updated in tandem.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -23,3 +23,6 @@
 	url = https://git.yoctoproject.org/poky
 	branch = langdale
 
+[submodule "meta-selinux"]
+	path = meta-selinux
+	url = https://git.yoctoproject.org/meta-selinux
--- a/meta-lxatac-software/conf/templates/default/bblayers.conf.sample
+++ b/meta-lxatac-software/conf/templates/default/bblayers.conf.sample
@@ -19,4 +19,5 @@ BBLAYERS ?= " \
   ##OEROOT##/../meta-labgrid \
   ##OEROOT##/../meta-rauc \
   ##OEROOT##/../meta-virtualization \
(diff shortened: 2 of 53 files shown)


hnez:
Forgot to mention @a3f in the original pull request description. Doing it now because I know he will be happy to see some progress on the `fastboot` front.

Emantor:
Why do we need this? Labgrid is perfectly fine with passing in the USB vendor id.

hnez:
Labgrid may be, but are our users as well?
At least I and @a3f were rather confused when `fastboot` proclaimed that there was no fastbootable device to be found.
The UX of the newer `fastboot` is better in that regard.

jluebbe:
> Why do we need this? Labgrid is perfectly fine with passing in the USB vendor id.

Note that since https://github.com/labgrid-project/labgrid/commit/5516d3629e558760f04d57c054a06f1cb4048422 we no longer pass the vendor ID, so this was broken on the LXA TAC without local patches.

a3f:
What's the blocker for merging this?

Emantor:
Passing the usb path still works, so I'm fine with this.

PR #37, opened 2023-04-06
Use a keyring directory for RAUC and add a devel certificate

hnez:
Adding a devel certificate that is deployed (but not enabled) on all LXA TAC images makes it easier for newcomers to build and install their first custom bundle.

The install-time detection of which certificates to trust is required for our planned release channels for official bundles:

First the bundles are built and signed with the `rauc-devel.key.pem` key and tested on hardware.
If the tests are passed the existing bundle will be automatically re-signed with the nightly release key.
If a new stable release is to be released an existing nightly bundle that has been manually tested is manually re-signed with the stable release key.

All of these bundles are deployed with the same `/etc/rauc/certificates-*` but a stable bundle should only be able to install other stable bundles, so we need to detect the release channel the bundle belongs to at install-time.


hnez:
We are about to generate certificates for our release channels. I think we should include those in this PR before merging it.

hnez:
The PR now contains the newly generated certificates and should be ready for merging.

meta-lxatac-bsp/recipes-core/rauc/rauc_%.bbappend
```diff
@@ -1 +1,50 @@
...
+    # of meta-lxatac this will point to devel.cert.pem and the file
+    # installed above is overwritten by this mv.
+    # If RAUC_KEYRING_FILE is overridden the extra cert will be installed
+    # along with the other ones.
+    mv ${D}${sysconfdir}/rauc/${KEYRING_FILE_NAME} \
+        ${D}${sysconfdir}/rauc/certificates-available/${KEYRING_FILE_NAME}
+
+    # Due to the certificate enable/disable logic in the RAUC hook the
+    # following line is only relevant for images _not_ installed via RAUC.
+    # Enable the certificate this bundle was signed with.
```
jluebbe: What do you mean by "this bundle", especially for images not installed via RAUC? Perhaps:
```suggestion
    # Enable the certificate this image was built with.
```

meta-lxatac-software/recipes-core/bundles/files/hook.sh
```diff
@@ -2,6 +2,39 @@
...
+			| openssl dgst -sha256 -c -hex \
+			| awk '{print toupper($2)}')
+
+		# Enable certificates that match the hash of the public key(s)
+		# that the current bundle is signed with.
+		# This means that a bundle signed with e.g. an official stable
+		# channel certificate will only be able to install other
+		# bundles from the same release channel.
+		for slot_hash in ${RAUC_BUNDLE_SPKI_HASHES}; do
+			if [ "${slot_hash}" == "${cert_hash}" ]; then
```
jluebbe: It's not a per *slot* hash, but for the bundle signature. Perhaps?
```suggestion
		for bundle_hash in ${RAUC_BUNDLE_SPKI_HASHES}; do
			if [ "${bundle_hash}" == "${cert_hash}" ]; then
```

jluebbe (changes requested):
Looks good otherwise.

Emantor (approved):

PR #39, opened 2023-05-11
meta-lxatac-bsp: barebox-tools: Update to release 2023.04.0

a3f:
This fixes a segmentation fault in imx-usb-loader experienced when uploading some types of i.MX6 images.


hnez:
Cool, thanks!

PR #40, opened 2023-05-15
meta-lxatac-software/labgrid: Export USB-Ports as Power Switches

SmithChart:
All three USB ports on the LXA TAC are power switchable (e.g. using `uhubctl`).
`labgrid` already has a driver that can handle ports where power can be switched: `USBPowerPort`.

With this change we'll export the three USB ports as power ports. This way they can be used to switch DUTs on or off.

The resources exported will look something like this:
```
$ labgrid-client r
lxatac-00013/dut_power/NetworkPowerPort
lxatac-00013/lxatac-usb-ports-power-p1/NetworkUSBPowerPort  # <-- new
lxatac-00013/lxatac-usb-ports-power-p2/NetworkUSBPowerPort  # <-- new
lxatac-00013/lxatac-usb-ports-power-p3/NetworkUSBPowerPort  # <-- new
lxatac-00013/serial/NetworkSerialPort
```

### DIFF
--- a/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
+++ b/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
@@ -44,6 +44,15 @@ dut_power:
     host: 'http://{{ hostname }}/v1/dut/powered/compat'
     index: '0'   # this is 'don't care'
 
+## Set up power switching for the USB ports on the LXA TAC
+{% for idx in range(1,4) %}
+lxatac-usb-power-p{{idx}}:
+  USBPowerPort:
+    match:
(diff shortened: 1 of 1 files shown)


hnez:
I've always assumed that we would implement this using the HTTP API of the `tacd` but it does not have to be that way.
Using `uhubctl` is perfectly fine, as it uses the same kernel API as the `tacd` to do the switching and the `tacd` notices the change eventually via [periodic polling](/linux-automation/tacd/blob/8232cf2d6b4cf66bd949214c2cde8a5a8dbb2fd8/src/usb_hub.rs#L196).

SmithChart:
Implementing this using the HTTP API shouldn't be any more difficult than using the ``USBPowerPort``. But a polling interval of _1s_ still feels quite responsive.

PR #41, opened 2023-05-23
meta-lxatac-bsp: barebox-tools: Update to release 2023.05.0

a3f:
This fixes imx-usb-loader -c wrongly reporting a verification failure because it reads beyond the barebox image.

The root case was the same underlying issue as what broke transmission in commit 44f60be228a0 ("meta-lxatac-bsp: barebox-tools: Update to release 2023.04.0"), but the verification path was only fixed in v2023.05.0.


hnez:
Neat, thanks!

PR #42, opened 2023-05-24
Do not clutter the rootfs's /srv on first boot

hnez:
This fixes an issue that comes up every time a LXA TAC is freshly deployed with a new eMMC image via fastboot (e.g. not via RAUC update).

On the first boot after deploying an image the partition that should be mounted to `/srv` is not yet generated by `systemd-repart`.
`systemd-repart` does however run later in the boot process than `systemd-gpt-auto-generator`.
As a consequence of that there is no `srv.mount` on first boot and every unit that writes to `/srv`  just clutters the rootfs with directories and files. This will result in issues on every subsequent boot as `systemd-gpt-auto-generator` does not generate mount units for directories that are not empty.

Currently we have to manually clear the cluttered `/srv/` after flashing a new image. This is not acceptable.

Instead use the `system-update.target` to perform just the repartioning on first boot and make sure no units that clutter `/srv` are pulled in by  `system-update.target` (e.g. via `sysinit.target`).

This PR supersedes #24 which tried to solve the same issue in a less fancy way. That PR can thus be closed.


michaelolbrich (approved):
Looks good to me.

Emantor (approved):

PR #45, opened 2023-06-07
linux-lxatac: update to kernel v6.4

hnez:
umpf tag is now: 6.4/customers/lxa/lxatac/20230607-1
The tag includes the following topic branches:

  - v6.4/customers/lxa/lxatac
  - v6.4/topic/lxa-tac-dts-mainlining
  - v6.4/topic/pwm-stm32
  - v6.4/topic/pwm-backlight

The devictree has been split out into its own topic branch to start its
mainlining journey.

The PR also contains a tweak to the `perf` recipe to build without `NO_LIBTRACEEVENT` to make it build with 6.4.

TODO before merging:

 - [x] Wait for 6.4 to be released and update this PR to use the released tarball.

Related Pull Requests
---------------

This PR depends on the following other PRs which require coordination:

- [x] The `tacd` PR that changes the expected regulator names to the ones used in the new devicetrees linux-automation/tacd#26.


hnez:
There is an [upstream patch](https://patchwork.yoctoproject.org/project/oe-core/patch/20230511125603.958144-4-max.oss.09@gmail.com/) for the perf/libtraceevent build issue. We should likely update the `perf.bbappend` with its content.

hnez:
I've looked into the upstream patch and have updated the `perf.bbappend` accordingly.
I've however decided to keep the `libtraceevent` PKGCONFIG off, as it requires a bit more research.
The time is better spent by just updating to a new upstream release that just ships with these changes once that's available.
We can live with the current solution until then.

Emantor (approved):

hnez:
I've added a commit that updates the tacd to a version with linux-automation/tacd#26 merged, rebased on top of the recent langdale branch and checked if everything works as expected.
It does, so I am going to merge it now …

PR #50, opened 2023-07-07
meta-labgrid: update to recent master

hnez:
This gives us an updated labgrid which, among other things fixes an issue that arises when using multiple USB-SD-Muxes or SD-Wires.

See labgrid-project/labgrid#1198 for more informations.


Bastian-Krause (approved):
Thanks!

Bastian-Krause:
> This gives us an updated labgrid which, among other things fixes an issue that arises when using multiple USB-SD-Muxes or SD-Wires.

Clarification: this can also happen with a single SD-Mux/SD-Wire. In this case the disk and control paths (obviously) cannot point to a different Mux, though.

PR #51, opened 2023-08-11
 lxatac-core-image-base: ship imx-uuu for uploading U-Boot

a3f:
We already ship barebox-tools which contains imx-usb-loader for uploading barebox binaries to i.MX SoCs, so let's build and ship uuu as well for uploading U-Boot binaries.

### DIFF
--- a/meta-lxatac-software/recipes-support/imx-uuu/imx-uuu_1.5.21.bb
+++ b/meta-lxatac-software/recipes-support/imx-uuu/imx-uuu_1.5.21.bb
@@ -1,0 +1,17 @@
+SUMMARY = "Universal Update Utility"
+DESCRIPTION = "Image deployment tool for i.MX chips"
+HOMEPAGE = "https://github.com/NXPmicro/mfgtools"
+
+SRC_URI = "https://github.com/nxp-imx/mfgtools/releases/download/uuu_${PV}/uuu_source-${PV}.tar.gz"
+SRC_URI[sha256sum] = "600be50827b52df4dddf0c7d07da27b103a4576eb445890905c61780e3c36871"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=38ec0c18112e9a92cffc4951661e85a5"
+
+inherit cmake pkgconfig
+
+S = "${WORKDIR}/uuu-${PV}"
+
+DEPENDS = "libusb zlib bzip2 openssl"
+
+BBCLASSEXTEND = "native nativesdk"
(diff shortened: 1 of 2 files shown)


meta-lxatac-software/recipes-support/imx-uuu/imx-uuu_1.5.21.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = "Universal Update Utility"
+DESCRIPTION = "Image deployment tool for i.MX chips"
+HOMEPAGE = "https://github.com/NXPmicro/mfgtools"
+
+SRC_URI = "https://github.com/nxp-imx/mfgtools/releases/download/uuu_${PV}/uuu_source-${PV}.tar.gz"
+SRC_URI[sha256sum] = "600be50827b52df4dddf0c7d07da27b103a4576eb445890905c61780e3c36871"
```
hnez, line 6: I don't know if relying on GitHub tarballs is still considered a good idea after [the incident](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/).
I don't see a strong commitment to keeping tarball hashes stable in the linked article.
@jluebbe what's your opinion on this?

jluebbe: Just fetch via git. This also reduces download size for future releases.

jluebbe: In this case, @a3f used the maintainer-uploaded .tar.gz, though. So we could also take this PR as-is.

jluebbe: In general, I'd prefer to just fetch via git. This also reduces download size for future releases.

In this case, @a3f used the maintainer-uploaded .tar.gz, though. So we could also take this PR as-is.

hnez: Ah, using uploaded assets was also a suggestion in the article.
I'll take the PR as-is then.

a3f: Thanks, I don't mind switching it to git if that's deemed better.

a3f: @jluebbe Do we need something else, so that users are able to force use of `uuu` when running `labgrid-client bootstrap`?

PR #54, opened 2023-09-05
meta-lxatac-software: move conf-notes.txt to template directory

ejoerns:
Otherwise it will not be found and used by `oe-setup-builddir` and user information about buildable targets is missing.

Fixes 27572845 ("meta-lxatac-software: move config templates to separate folder")


hnez:
Thanks! I never noticed that the conf notes went missing. Having them back should be quite helpful to new users.

PR #55, opened 2023-09-20
oe-init-build-env: make script independent of `pwd`

userid0x0:
Hi,
this is only a slight change in the `oe-init-build-env` that removes the requirement to source the script from `meta-lxatac`.
Thus it works fine with crops/yocto containers:
```bash
# Podman
podman run --rm --tty --interactive \
  --userns=keep-id --user $(id -u):$(id -g) \
  --env HOME=/tmp --tmpfs /tmp \
  --volume $(pwd)/meta-lxatac:$(pwd)/meta-lxatac \
  ghcr.io/crops/yocto:debian-10-builder \
  --pokydir $(pwd)/meta-lxatac --target lxatac-core-bundle-base --builddir $(pwd)/meta-lxatac/build
```

### DIFF
--- a/oe-init-build-env
+++ b/oe-init-build-env
@@ -11,9 +11,12 @@ else
     echo "  exec bash -c \"source oe-init-build-env && exec fish\""
     exit 1
 fi
+
 DIR=$( dirname -- "${THIS_SCRIPT}" )
+DIR=$( readlink -f "${DIR}" )
 unset THIS_SCRIPT
 
-TEMPLATECONF=${DIR}/meta-lxatac-software/conf/templates/default/
+TEMPLATECONF="${DIR}/meta-lxatac-software/conf/templates/default/"
 
-. ${DIR}/poky/oe-init-build-env $1
+# shellcheck source=poky/oe-init-build-env
+. "${DIR}/poky/oe-init-build-env" "$1"
--- a/oe-init-build-env
+++ b/oe-init-build-env
@@ -1,5 +1,19 @@
 #!/usr/bin/env bash
 
-TEMPLATECONF=`pwd`/meta-lxatac-software/conf/templates/default/
+if [ -n "${BASH_SOURCE[0]}" ]; then
+    THIS_SCRIPT="${BASH_SOURCE[0]}"
+elif [ -n "$ZSH_NAME" ]; then
+    THIS_SCRIPT="$0"
+else
+    echo "The meta-lxatac oe-init-build-env can only be sourced by bash and zsh."
+    echo "For other shells (e.g. fish) use a workaround like the following:"
+    echo
+    echo "  exec bash -c \"source oe-init-build-env && exec fish\""
+    exit 1
+fi
+DIR=$( dirname -- "${THIS_SCRIPT}" )
+unset THIS_SCRIPT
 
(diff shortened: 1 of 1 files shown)


oe-init-build-env
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
+DIR=$( dirname -- "${BASH_SOURCE[0]}" )
+TEMPLATECONF=${DIR}/meta-lxatac-software/conf/templates/default/
```
hnez: Hi,

our first external contribution, exciting!

The previous solution using `pwd` made sure that `$TEMPLATECONF` is an absolute path.
Now `$TEMPLATECONF` can be a relative path (e.g. `./`). This results in the script failing to source (at least for me):

```
lgo@dude03:~/work/Projects/meta-lxatac$ source oe-init-build-env
Error: TEMPLATECONF value points to nonexistent directory './meta-lxatac-software/conf/templates/default/'
```

I would suggest taking inspiration from `poky/oe-init-build-env` and adding a `readlink -f`:

https://github.com/yoctoproject/poky/blob/5614e6b38a55df5df95556853961610c2b6bf329/oe-init-build-env#L38-L42

```suggestion
DIR=$( dirname -- "${BASH_SOURCE[0]}" )
DIR=$( readlink -f "${DIR}" )

TEMPLATECONF=${DIR}/meta-lxatac-software/conf/templates/default/
```

userid0x0: Not a big deal. I integrated your suggestion into the commit and force-pushed.

oe-init-build-env
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
+DIR=$( dirname -- "${BASH_SOURCE[0]}" )
```
hnez: Now that I've had a second look I think we should actually copy the `zsh` compatibility code from poky, as `zsh` is kind of widespread:

```suggestion

if [ -n "${BASH_SOURCE[0]}" ]; then
    THIS_SCRIPT="${BASH_SOURCE[0]}"
elif [ -n "$ZSH_NAME" ]; then
    THIS_SCRIPT="$0"
else
    echo "The meta-lxatac oe-init-build-env can only be sourced by bash and zsh."
    echo "For other shells (e.g. fish) use a workaround like the following:"
    echo
    echo "  exec bash -c \"source oe-init-build-env && exec fish\""
    exit 1
fi

DIR=$( dirname -- "${THIS_SCRIPT}" )
unset THIS_SCRIPT

```

hnez (changes requested):
Could you also copy the motivation for this change (e.g. basically the pull request comment) to the git commit description?
This makes browsing the history outside of github easier.

hnez (approved):

hnez:
The `readlink -f "${DIR}"` went missing with the latest change. I've added a commit that re-introduces it and pushed it to your branch. I hope that's okay for you.

I also ran [shellcheck](https://www.shellcheck.net/) on the script (because I don't trust my shell-scripting skills) and fixed warnings regarding unquoted use of variables.

The PR now looks good to me, so I am going to click merge.
Thanks for your contribution!

userid0x0:
You are correct, I wanted to do a negative test as well.
Thanks for your assistance here.

PR #56, opened 2023-09-22
README: fix ssh-keys.bb example

userid0x0:
I think the `file://` prefix is missing.


hnez:
Yes, you are right. Thanks!

PR #61, opened 2023-09-27
meta-lxatac-software: tacd{,-webinterface}: dependency updates

hnez:
Update the tacd and its web interface

TODO
----

- [x] Switch to a newer `rustc` version to get around:
  ```
  error: package `zvariant_derive v3.15.0` cannot be built because
  it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
  ```
  The version in `mickledore` is already new enough, so we can just wait this out.

Related Pull Requests
---------------

This PR is related to another PRs that should be coordinated:

- [x] The `tacd` pull request that does the dependency update there: linux-automation/tacd#42


hnez:
Assigned me to rebase, update and change the target branch once #70 is in.

hnez:
I've updated the recipe to point to the most recent `tacd` and made sure it builds on mickledore's `rustc`.
It built just fine, so I think we are ready to merge.

jluebbe (approved):

PR #62, opened 2023-10-10
emmc-image: remove size workaround

rohieb:
genimage knows about the 'size' option for images to fill the image to a certain size, and since genimage v14 it supports the 'fill' option to write all zeroes at the end of the image and not cut them off to optimise image size. Use this to blow up the images to the full 2 GiB, and remove our invisible-partition-at-the-end hack.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
+++ b/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
@@ -8,7 +8,7 @@ image @IMAGE@img {
     size = 2G
     hdimage {
         align = 1M
-        gpt = "true"
+        partition-table-type = gpt
         fill = true
     }
 
--- a/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
+++ b/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
@@ -5,9 +5,11 @@ image @IMAGE@simg {
 }
 
 image @IMAGE@img {
+    size = 2G
     hdimage {
         align = 1M
         gpt = "true"
+        fill = true
     }
 
     partition reserved {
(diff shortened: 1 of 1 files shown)


hnez:
When building with this patch applied i get:

1) The following info (unrelated to this PR, but this would be a great opportunity to fix it as well):

        INFO: hdimage(emmc-image-lxatac-20231010135745.img): The option 'gpt' is deprecated.
        Use 'partition-table-type' instead

2) The following error:

        ERROR: hdimage(emmc-image-lxatac-20231010135745.img): partition [GPT backup]
        (offset 0x7fffbe00, size 0x4200) overlaps previous partition root-a (offset 0x200000, size 0x80000000)

I did not yet look into the cause of the error, so I'll leave this as a reminder to look into it.

rohieb:
> When building with this patch applied i get:
> 
>     1. The following info (unrelated to this PR, but this would be a great opportunity to fix it as well):
>        ```
>         INFO: hdimage(emmc-image-lxatac-20231010135745.img): The option 'gpt' is deprecated.
>         Use 'partition-table-type' instead
>        ```

Ah yes, I've included a fix for this warning too.

>     2. The following error:
>        ```
>         ERROR: hdimage(emmc-image-lxatac-20231010135745.img): partition [GPT backup]
>         (offset 0x7fffbe00, size 0x4200) overlaps previous partition root-a (offset 0x200000, size 0x80000000)
>        ```

I should have forseen that – the 2048 MiB root partition no longer fits into the 2 GiB image when there is a GPT primary header (with protective MBR) at the start and a GPT backup header at the end of the image, and a 1 MiB reserved partition. I've now also decreased the rootfs size to 2045 MiB so it fits again.

hnez (approved):
I've worked on the labgrid testsuit for the LXA TAC the last few days so that it now also checks some aspects of the file system and partitions on the running system.
This PR passed these new tests, so they look good to me!

jluebbe:
Isn't a rootfs-a size of 2045MiB inconsistent with https://github.com/linux-automation/meta-lxatac/blob/langdale/meta-lxatac-bsp/recipes-core/lxatac-repart/files/repart.d/50-root-a.conf? Does this trigger a resize of the active rootfs partition on first boot?

I'd expect the image to be extended so that a 2GiB rootfs fits.

rohieb:
Hmmm, as that file should only be used for matching the root partition, we could even remove the `SizeMinBytes` and `SizeMaxBytes` settings, so systemd-repart does not touch the first root partition. `Type` and `Label` should be enough for matching it.

PR #64, opened 2023-10-19
GitHub/GitLab Action/Pipeline Runners

hnez:
This allows registering the TAC itself as a GitHub/GitLab runner. Once registered the TAC can be used to flash artifacts onto a DUT and execute tests.

Having this feature makes the TAC a lot more useful (in my opinion) but one has to keep in mind that using it makes the TAC effectively a backdoor into your infrastructure for whoever runs you GitHub/GitLab infrastructure.
I have some ideas on how to mitigate this a bit (run the runner services in network namespaces and as normal users with adequate permissions and firewalling), but these are not implemented in this PR.

Handle with care

TODO
----

- [x] Add service files (disabled by default) to start the runners.
  - [x] GitHub Runner
  - [x] GitLab Runner
- [x] Play around a bit more with the runners.
- [x] Investigate the build time and image size impact of this PR. The go binaries feel a bit heavy.
  - The build time increase of the Go binaries is not that bad, as we already have a Go toolchain for e.g. podman.
  - The build time increase for the target nodejs is a bit annoying though
  - The size of the RAUC bundle increased from `229M` to `268M` (`39M` or 17% larger than before). Not great, but for the feature well worth it, I guess.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/gitlab-runner/gitlab-runner.inc
+++ b/meta-lxatac-software/recipes-devtools/gitlab-runner/gitlab-runner.inc
@@ -1,0 +1,30 @@
+SUMMARY = "GitLab Runner"
+DESCRIPTION = "The Runner for GitLab Pipelines"
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI:append = " file://gitlab-runner.service "
+
+GO_IMPORT = "gitlab.com/gitlab-org/gitlab-runner"
+GO_INSTALL = "${GO_IMPORT}"
+
+RDEPENDS:gitlab-runner-dev = "bash"
+
+inherit go-mod
+inherit systemd
+
+SYSTEMD_SERVICE:${PN} = "gitlab-runner.service"
+
+# This is required because the go build system fetches dependecies in the
+# compile stage.
+do_compile[network] = "1"
+
+# This is required to prevent build failures due to
+# "duplicated definition of symbol" errors.
+GO_LINKSHARED = ""
+
+do_install:append() {
+    install -d ${D}${systemd_system_unitdir}
+    install -m 0644 ${WORKDIR}/gitlab-runner.service ${D}${systemd_system_unitdir}/
+}
--- a/meta-lxatac-software/recipes-core/bundles/files/hook.sh
+++ b/meta-lxatac-software/recipes-core/bundles/files/hook.sh
@@ -75,6 +75,7 @@ case "$1" in
 		migrate /etc/labgrid/userconfig.yaml
 		migrate /etc/github-act-runner/sessions.json
 		migrate /etc/github-act-runner/settings.json
+		migrate /etc/gitlab-runner/config.toml
 		for x in /etc/ssh/ssh_host*; do
 			migrate "${x}"
 		done
--- a/meta-lxatac-software/recipes-core/bundles/files/hook.sh
+++ b/meta-lxatac-software/recipes-core/bundles/files/hook.sh
@@ -73,6 +73,8 @@ case "$1" in
 		migrate /etc/machine-id
 		migrate /etc/labgrid/environment
 		migrate /etc/labgrid/userconfig.yaml
+		migrate /etc/github-act-runner/sessions.json
+		migrate /etc/github-act-runner/settings.json
 		for x in /etc/ssh/ssh_host*; do
 			migrate "${x}"
 		done
--- a/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
+++ b/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
@@ -46,6 +46,7 @@ IMAGE_INSTALL:append = "\
     gdbserver \
     git \
     github-act-runner \
+    gitlab-runner \
     gstreamer1.0 \
     gstreamer1.0-plugins-bad-videoparsersbad \
     gstreamer1.0-plugins-base \
--- a/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
+++ b/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
@@ -45,6 +45,7 @@ IMAGE_INSTALL:append = "\
     gdb \
     gdbserver \
     git \
+    github-act-runner \
     gstreamer1.0 \
     gstreamer1.0-plugins-bad-videoparsersbad \
     gstreamer1.0-plugins-base \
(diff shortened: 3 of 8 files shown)


meta-lxatac-software/recipes-devtools/gitlab-runner/gitlab-runner.inc
```diff
@@ -0,0 +1,29 @@
...
+inherit go
+inherit go-mod
+inherit systemd
+
+SYSTEMD_SERVICE:${PN} = "gitlab-runner.service"
+
+do_compile() {
+    export TMPDIR="${GOTMPDIR}"
+    ${GO} install ${GOBUILDFLAGS}
+}
```
jluebbe: This needs a comment why the default from the go classes is not sufficient.

hnez: Oh yeah, I kind of forgot that I had to add that. I've updated the commit and moved the workaround a bit closer to the actual culprit:

https://github.com/linux-automation/meta-lxatac/blob/755c3c23c9399ad2e591bf413ab30666caa1a36c/meta-lxatac-software/recipes-devtools/gitlab-runner/gitlab-runner.inc#L21-L30

I don't really have a good answer as to _why_ the compile stage fails, because I don't know to much about Go and how to package Go projects.

Here is an excerpt from the do_compile logs:

```
DEBUG: Executing shell function do_compile
HASH[moduleIndex]
HASH[moduleIndex]: "go1.19.7"
HASH[moduleIndex]: "modroot /ptx/work/WORK_SANAH/lgo/Projects/meta-lxatac/build/tmp/work/cortexa7t2hf-neon-vfpv4-oe-linux-gnueabi/gitlab-runner/16.2.2-r0/build/src/gitlab.com/gitlab-org/gitlab-runner\n"
…
HASH[moduleIndex]: "file single.go 2023-10-23 12:18:37.248464983 +0000 UTC 5210\n"
HASH[moduleIndex]: "file single_test.go 2023-10-23 12:18:37.248464983 +0000 UTC 3289\n"
HASH[moduleIndex]: "file Vagrantfile 2023-10-23 12:18:37.236464946 +0000 UTC 1986\n"
HASH[moduleIndex]: "file config.toml.example 2023-10-23 12:18:37.256465005 +0000 UTC 669\n"
HASH[moduleIndex]: "file unregister.go 2023-10-23 12:18:37.248464983 +0000 UTC 2604\n"
... 50.000 similar lines ...
HASH[link gitlab.com/gitlab-org/gitlab-runner]: 51391287bfd80a3cf807d8f80fa8d621e6d0b02c9a095de9da62b2afc6b68609
HASH subkey 4eba7844681da8d69748f63e968ed43b568face86d54d7a118abda3a51623ff5 "link-stdout" = c4e9afa14a7d97bef211d92e76f7d06c0e608a848d8d54e01528885ad669bb53
HASH subkey 6b98acb0b9b3698479e89ec43fe407ed39e0972fe600d46b6456c6cedc82226f "link-stdout" = 6b84cae771da7444fd35ad8e010da0dc707687a1b07a4bce58c1530f04f37617
HASH subkey 51806134918812b94be74eb1507758c70ec7d2c4abaa90cd3a121a6bf2a7fe98 "link-stdout" = f4db7eb852d9f0b26c3cc3a97f1b834a6a7a3bed0c7b86b0e71a8b02548c1125
HASH subkey 5a95087a85e736d91c9245885f3e51a83c14fdbae73d12e98ecc6356b69e8476 "link-stdout" = 64c050825db8d5d4a74d35aabc117f02c62399fe873178b1633026f6bbe9212e
HASH subkey 80cda6fa8385d2022e4a2cca1cafab81bfd235b4b6a0244b41842eec5e938072 "link-stdout" = 43db68b16b5720db3c90e10490d3ee19a204e1962ca880de5d660c128688f553
HASH subkey 39a95685e237b046cf7bb71e6dc2b8118ee450b7fabde6ebc0d966efcf2a7b9e "link-stdout" = effaa42e9e940d9b2b22326f658cfcd1db9ff2e9fec0086f4df1e1a7d551d0ea
HASH subkey 7a01f21523fee0823f613cb5be41074b3f0afa10ae82bd9a124c8cb03030d487 "link-stdout" = 74150482e2cd43588d973ede4ead023f349525e2d1f1f57b195206e9c816aa66
WARNING: exit code 2 from a shell command.
```

The `go.bbclass` tries to pass all these lines as arguments to `go install` (as far as I can tell).

hnez:
Assigned myself because I need to take a second look at the go packaging situation.

meta-lxatac-software/recipes-devtools/gitlab-runner/gitlab-runner.inc
hnez: I was wrong here. All of the verbose `HASH[…]` lines were output that `go list` outputs on stderr, so these did not get piped anywhere.

The underlying problem was that the `go.bbclass` does by default compile with the `-linkshared` flag, which for the gitlab-runner fails due to duplicate symbols in the generated binary.
When compiling without `-linkshared` the duplicate symbols do not seem to be an issue.
 
By setting `GO_LINKSHARED = ""` the `-linkshared` flag is omitted and the recipe builds without the ugly hack.

hnez:
Apart from requiring `do_compile[network] = "1"` the recipes are in a good shape now, I think.
And having that in go recipes seems to be common enough that even [meta-oe](https://github.com/openembedded/meta-openembedded/blob/d43a66fdc57a68faf54bf8f27d1d2f83ed452cdc/meta-oe/recipes-support/crucible/crucible_2023.11.02.bb) does it.

Here are two test runs …

- … one on [github.com](https://github.com/hnez/usbsdmux/actions/runs/7298965495/job/19890887467)
- … and one on [gitlab.com](https://gitlab.com/hnez/lxatac-runner-test/-/jobs/5809462180)

I've quickly de-registered the runners again though, because of the security implications.

jluebbe:
GitHub seems to expect that the runner can update itself: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners

I've seen that outdated runners are rejected relatively soon (weeks), so I'm not sure how useful it is to ship the runner in the image. What do you think?

hnez:
Hi @jluebbe,

I've had a look at the `github-act-runner` commit history and to me it does not look like they are playing a constant game of cat and mouse against the GitHub API.
They have a [CI job that tests against regression](https://github.com/ChristopherHX/github-act-runner/actions?page=9) which does not seem to have caught a regression since its introduction six months ago.

I don't think we should advertise too much with the GitHub action runner feature, because it could stop working at any time, as you have said. It uses an unofficial client, that could just stop being developed and I have only tested a small portion of its features (for example [job.<job_id>.container](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container) will most likely not work because we do not have a docker daemon running).
But I would like to experiment with GitHub/GitLab runners and that's a lot easier when they are part of the images we publish, so I do not have to build my own.

jluebbe (approved):

PR #65, opened 2023-10-23
meta-lxatac-bsp: rauc: resize partitions after writing them

hnez:
We've discovered that LXA TACs only filled the whole root parition with a filesystem on the initial deploy, but not for subsequent RAUC updates. As it turns out we were missing the resize flag in system.conf.

Fix that.

TODO before merging:

- [x] ~~The current bundles do not ship a `resize2fs`, making installation of bundles built with this impossible. Find and implement a workaround (like installing the `resize2fs` shipped in the bundle in a hook)~~ Nope. This was just me being confused. The PR works just fine as-is.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/rauc/files/lxatac/system.conf
+++ b/meta-lxatac-bsp/recipes-core/rauc/files/lxatac/system.conf
@@ -11,11 +11,13 @@ directory=certificates-enabled
 device=/dev/disk/by-partuuid/e82e6873-62cc-46fb-90f0-3e936743fa62
 type=ext4
 bootname=system0
+resize=true
 
 [slot.rootfs.1]
 device=/dev/disk/by-partuuid/8eb3e87e-2b4e-45e3-888b-2f678662862d
 type=ext4
 bootname=system1
+resize=true
 
 [slot.bootloader.0]
 device=/dev/mmcblk1
--- a/meta-lxatac-bsp/recipes-core/rauc/rauc_%.bbappend
+++ b/meta-lxatac-bsp/recipes-core/rauc/rauc_%.bbappend
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-RDEPENDS:${PN}:append = "bash"
+RDEPENDS:${PN}:append = "bash e2fsprogs-resize2fs"
 
 SRC_URI:append = " \
     file://require-mount-srv.conf \


meta-lxatac-software/recipes-core/bundles/files/hook.sh
```diff
@@ -44,7 +44,22 @@ function migrate () {
...
+		echo "#!/bin/sh" > /sbin/resize2fs
+		echo "echo \"Dummy resize2fs installed by RAUC hook\"" >> /sbin/resize2fs
+		chmod +x /sbin/resize2fs
+	fi
+}
+
 case "$1" in
+	slot-pre-install)
+		mock_resize2fs
+		;;
```
hnez: This does not work for us as RAUC tries to mount the old slot before running the `pre-install` hook.
On freshly deployed TACs there is however no partition to mount and the installation fails 😒.

hnez: Nevermind. I misunderstood how the settings in in `system.conf` affect the generated bundle (_not at all_).
The generated bundles install just fine on existing devices (without resizing the installed rootfs though) and subsequent installs will have a rootfs that is grown to fit.

meta-lxatac-software/recipes-core/bundles/lxatac-core-bundle-base.bb
```diff
@@ -18,7 +18,7 @@ RAUC_BUNDLE_SLOTS ?= "rootfs bootloader"
 
 RAUC_SLOT_rootfs ?= "lxatac-core-image-base"
 RAUC_SLOT_rootfs[fstype] = "ext4"
-RAUC_SLOT_rootfs[hooks] = "post-install"
+RAUC_SLOT_rootfs[hooks] = "pre-install;post-install"
```
jluebbe: Drop this change.

meta-lxatac-software/recipes-core/bundles/files/hook.sh
jluebbe: Drop this change and update the commit message.

jluebbe (changes requested):

meta-lxatac-software/recipes-core/bundles/files/hook.sh
hnez: Ooops, looks like I've opened the pull request from a branch here and not from by fork.
I've updated the branch here now.

jluebbe (approved):

PR #67, opened 2023-10-25
meta-lxatac-software: labgrid-exporter: more specific usb serial match

### LINKED ISSUE #66: Too broad matching of USBSerialPorts
When exporting separate USBSerialPorts on the same USB port, the match of the `lxatac-usb-ports-px` USBSerialPort should be `'{{sysfs}}:1.0'` instead of `'{{sysfs}}'`. Otherwise udev events might confuse labgrid leading to the same ttyUSBx being used for both `lxatac-usb-ports-px` and `lxatac-usb-ports-px.y`.
https://github.com/linux-automation/meta-lxatac/blob/6b6cbb54276bfbc4152603fab02a4ee7ed7f28cd/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml#L89-L102

hnez:
The less-specific `{{sysfs}}` match also matches on `{{sysfs}}:1.1`, `{{sysfs}}:1.2`, `{{sysfs}}:1.3` etc..
This results in labgrid being rightfully confused about udev events for said devices.

Match on the more specific `{{sysfs}}:1.0` instead.

A side-effect of this change is that the first port is now exported as `lxatac-usb-ports-p1.0` instead of `lxatac-usb-ports-p1`.

This fixes #66.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
+++ b/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
@@ -88,7 +88,7 @@ lxatac-usb-ports-p{{idx}}:
       '@ID_PATH': '{{sysfs}}'
   USBSerialPort:
     match:
-      '@ID_PATH': '{{sysfs}}'
+      '@ID_PATH': '{{sysfs}}:1.0'
   USBVideo:
     match:
       '@ID_PATH': '{{sysfs}}'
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Looks okay from the configuration side, but it will break a lot of existing place matches. Is it worth it?

jluebbe:
To keep compatibility, we could just add `:1.0` to the "default" match under `lxatac-usb-ports-p1`.

hnez:
> add `:1.0` to the "default" match under `lxatac-usb-ports-p1`.

That's okay with me. Did that just now.

jluebbe (approved):

jluebbe:
@Bastian-Krause Are you fine with this?

Bastian-Krause (approved):

PR #68, opened 2023-10-27
meta-lxatac-software: android-tools: do not use host /usr in libext4_utils

hnez:
Without this patch the build rightfully fails with:

```
cc1plus: error: include location "/usr/include/android" is unsafe for cross-compilation [-Werror=poison-system-directories]
```

This issue was discovered by @ukleinek when building the BSP from scratch, which we do not usually do (but maybe should do to discover these kinds of issues).


jluebbe (approved):
Seems plausible.

PR #69, opened 2023-10-27
meta-lxatac-bsp: emmc-image: build smaller images (faster)

hnez:
This PR does three things:

- Revert #62 by @rohieb because it did not achieve quite what we wanted
- Remove the `partition end` workaround that was removed in #62 (but without changing other things like adding `fill=true`).
- Do not place a GPT header at the "end"¹ of the image.
- Use the existing rootfs ext4 instead of generating a new one from the `.tar`.

Things I am not too sure about:

- [x] Should we use some templating magic instead of writing `lxatac-core-image-base-lxatac.ext4` in the `genimage.config`?
- [x] Are all the dependencies automatically set up so that we can rely on `lxatac-core-image-base-lxatac.ext4` being present.
     The first version of this PR used `GENIMAGE_ROOTFS_IMAGE` to specify this dependency. This is not quite correct. The `genimage.bbclass`  documentation states quite clearly that this variable is intended for `.tar` images that should be unpacked and re-packed into a filesystem image (thx @ejoerns ). Instead I've now set up the dependencies explicitly.
- [x] Is the rootfs grown like we want it to.

The last one I can test, for the rest I would be glad about some feedback.

---

¹ Which is the end of the image but has nothing to do with the end of the _device_ the image is written to, where the header should be.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/images/emmc-image.bb
+++ b/meta-lxatac-bsp/recipes-core/images/emmc-image.bb
@@ -12,9 +12,7 @@ inherit genimage
 COMPATIBLE_MACHINE = "lxatac"
 
 GENIMAGE_IMAGE_SUFFIX = ""
-GENIMAGE_ROOTFS_IMAGE = "lxatac-core-image-base"
-GENIMAGE_ROOTFS_IMAGE_FSTYPE = "tar"
 
 do_genimage[depends] += " \
-    ${GENIMAGE_ROOTFS_IMAGE}:do_image_complete \
+    lxatac-core-image-base:do_image_complete \
 "
--- a/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
+++ b/meta-lxatac-bsp/recipes-core/images/emmc-image/lxatac/genimage.config
@@ -5,11 +5,9 @@ image @IMAGE@simg {
 }
 
 image @IMAGE@img {
-    size = 2G
     hdimage {
         align = 1M
         partition-table-type = gpt
-        fill = true
     }
 
     partition reserved {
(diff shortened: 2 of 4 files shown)


rohieb:
The seond one you could also check by inspecting the `DEPENDS` variable in the output of `bitbake -e emmc-image`.

Edit: Ah, the `do_genimage[depends] += "${GENIMAGE_ROOTFS_IMAGE}:do_image_complete"` should do the trick, as long as the `do_image_complete` task in *lxatac-core-image-base* creates the ext4 image.

meta-lxatac-bsp/recipes-core/images/emmc-image.bb
```diff
@@ -12,9 +12,7 @@ inherit genimage
 COMPATIBLE_MACHINE = "lxatac"
 
 GENIMAGE_IMAGE_SUFFIX = ""
-GENIMAGE_ROOTFS_IMAGE = "lxatac-core-image-base"
-GENIMAGE_ROOTFS_IMAGE_FSTYPE = "tar"
 
 do_genimage[depends] += " \
-    ${GENIMAGE_ROOTFS_IMAGE}:do_image_complete \
+    lxatac-core-image-base:do_image_complete \
 "
```
hnez, line 18: I've changed how the dependency on the rootfs image is handled because `GENIMAGE_ROOTFS_IMAGE` is no longer set.
See the updated PR description for more detail.

hnez:
Seeing the `core` and `base` in the image name always irks me a little bit, as it is not very core-y or base-y with all the development tools.
We should define an actual minimal image in the future and derive our full image with all the tools from there.
(But that's not something we should do or discuss here)

jluebbe (approved):

meta-lxatac-bsp/recipes-core/images/emmc-image.bb
jluebbe: Looks good.

hnez:
As it turns out the rootfs was not resized to match the partition as we wanted it to.
Newer `systemd` versions ship a `systemd-growfs-root.service` that can help us here, so I've backported it for now.
Hopefully we can throw away the backport very soon.

jluebbe (approved):

PR #70, opened 2023-10-27
Update to Yocto 4.2 - Mickledore

hnez:
This is the long-awaited update from Langdale (4.1) to Mickledore (4.2)

Things to do before merging¹:

- [x] Fix the underlying issue that makes `trace-cmd` fail to build and fix it instead of disabling the package.
  2023-11-01: It looks like this issue fixed itself in the meantime 🎉.
- [x] `systemd-repart.service` fails to start with "Could not determine temporary directory: No such file or directory", which needs to be fixed.
- [x] `barebox-state` does not work. This can be observed by running e.g. `rauc status mark-bad other`:
    `rauc mark: Failed marking slot rootfs.1 as bad: barebox backend: Failed to run barebox-state: Child process exited with code 1`
    This makes RAUC unusable, which needs to be fixed.
- [x] Test on hardware

Other Pull requests that need coordination:

- [ ] #61 can not happen before this is merged, so it needs to be rebased on top of this work and the target branch has to be updated.
- [x] #69 backports a `systemd` service file. This should become obsolete in mickledore, so the backport should be removed again.
    2023-11-01: added a commit that removes the file again.

---

¹ Due to the way meta layers are ususually handled, e.g. having a branch name matching the release name I have created a mickledore branch based on the current langdale branch content to be able to open this PR against it.


meta-lxatac-software/recipes-rust/bottom/bottom_0.6.8.bb
```diff
@@ -182,7 +182,170 @@ SRC_URI += " \
     crate://crates.io/winapi/0.3.9 \
 "
 
-
+SRC_URI[addr2line-0.16.0.sha256sum] = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
```
jluebbe, line 185: Could you add the command you used for the update to the commit message?

hnez: Sure!
The `cargo.bbclass` makes the process quite straight forward. Just build the recipe without the hashes, then `cargo.bbclass` will make the build fail and print out a list of hashes to copy and paste.

hnez:
It looks like the newer `barebox-state` looks for the state on one of the eMMC hardware boot partitions:

```
$ barebox-state --version
dt-utils 2023.08.0
Configured with build-time option '--disable-state-backward-compatibility'.
                                  '--disable-lock-device-node'.
$ strace barebox-state
…
openat(AT_FDCWD, "/dev/mmcblk1boot1", O_RDONLY|O_LARGEFILE) = 4
ioctl(4, MIXER_READ(1) or MEMGETINFO, 0xbefe8ae8) = -1 EINVAL (Invalid argument)
close(4)                                = 0
openat(AT_FDCWD, "/dev/mmcblk1boot1", O_RDONLY|O_LARGEFILE) = 4
openat(AT_FDCWD, "/dev/mmcblk1boot1", O_RDONLY|O_LARGEFILE) = 5
openat(AT_FDCWD, "/dev/mmcblk1boot1", O_RDONLY|O_LARGEFILE) = 6
_llseek(4, 786432, [786432], SEEK_SET)  = 0
read(4, "\271&\3\0\r-\3\0", 8)          = 8
write(2, "state: No meta data header found"..., 33state: No meta data header found
) = 33
_llseek(5, 790528, [790528], SEEK_SET)  = 0
read(5, "\10\20\7\26\34P\0\0", 8)       = 8
write(2, "state: No meta data header found"..., 33state: No meta data header found
) = 33
_llseek(6, 794624, [794624], SEEK_SET)  = 0
```

While the old `barebox-state` finds it on the normal eMMC partition:

```
$ barebox-state --version
dt-utils 2021.03.0
Configured with build-time option '--disable-state-backward-compatibility'.
$ strace barebox-state
openat(AT_FDCWD, "/dev/mmcblk1", O_RDONLY|O_LARGEFILE) = 4
ioctl(4, MIXER_READ(1) or MEMGETINFO, 0xbefcaaa0) = -1 EINVAL (Invalid argument)
close(4)                                = 0
openat(AT_FDCWD, "/dev/mmcblk1", O_RDWR|O_LARGEFILE) = 4
openat(AT_FDCWD, "/dev/mmcblk1", O_RDWR|O_LARGEFILE) = 5
openat(AT_FDCWD, "/dev/mmcblk1", O_RDWR|O_LARGEFILE) = 6
_llseek(4, 786432, [786432], SEEK_SET)  = 0
```

In my opinion having the state on the normal partition makes more sense, as the active boot partition switches with every RAUC update. I did not yet investigate further @a3f have you seen this before?

a3f:
> In my opinion having the state on the normal partition makes more sense, as the active boot partition switches with every RAUC update. 

Agreed.

> I did not yet investigate further @a3f have you seen this before?

dt-utils iterates via libudev over all block devices in cdev_from_block_device to find the matching device. Indeed all of boot0, boot1 and the user area have type "disk", so barebox-state gets the wrong device thinking it's the disk device. I am not sure how to best differentiate between boot and user partitions. Maybe check prefer block devices that have no block devices as a parent?

hnez:
I've worked around the `barebox-state` issue for now by stepping the `meta-rauc` submodule back by [one commit](https://github.com/rauc/meta-rauc/commit/c0dcec7b1b0d20862806915e12b4f332e5bef58c) (the one upgrading to `dt-utils` 2023.08.0).
This is obviously not a permanent solution, but combined with another workaround for `systemd-repart` I can now at least pass the (not yet public - sorry potential outside readers) LXA TAC labgrid testsuite.

I think I will do some manual testing and then mark this PR as ready for review.
Would be glad to take suggestions for more elegant solutions for the aforementioned workarounds.

a3f:
> dt-utils iterates via libudev over all block devices in cdev_from_block_device to find the matching device. Indeed all of boot0, boot1 and the user area have type "disk", so barebox-state gets the wrong device thinking it's the disk device. I am not sure how to best differentiate between boot and user partitions. Maybe check prefer block devices that have no block devices as a parent?

For now, I restored the old behavior, namely taking the first "disk" again instead of the last. Fix is here: https://lore.barebox.org/oss-tools/20231106110405.3209233-1-a.fatoum@pengutronix.de/T/#u. IMO, it's worth a bugfix release, so once tested, we can kick that off.

hnez:
I've included the `barebox-state` patch and tested manually and using the LXA TAC testsuite.
It looks to be working as intended.

With this I will mark this PR as "Ready for review".

**PS:** Something weird is going on with GitHubs commit list. It shows commits before cc4d0ea as part of this PR, even though they are already in the "`mickledore`" branch (which matches current `langdale`).

hnez:
Now that #71 is merged we should have everything we want to have in a first stable release (not having it would have been annoying because we would have needed to keep the workaround documentation around somewhere).

This means we should leave langdale behind and concentrate on getting this PR merged.
All other open PRs should then be updated to go to the mickledore branch instead.

jluebbe:
> **PS:** Something weird is going on with GitHubs commit list. It shows commits before [cc4d0ea](https://github.com/linux-automation/meta-lxatac/commit/cc4d0eabc4215667e2f18857b559dc8fc333b1f8) as part of this PR, even though they are already in the "`mickledore`" branch (which matches current `langdale`).

I've toggled the base branch to `langdale` and back to `mickledore`, which seems to have repaired the list of changes to review.

meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
```diff
@@ -9,6 +9,10 @@ IMAGE_FSTYPES += "ext4"
 # use a fixed directory hash seed to reduce the image delta size
 EXTRA_IMAGECMD:ext4 = "-i 4096 -E hash_seed=4267a9cf-754d-4506-9156-d3f4a18842e5"
 
+# Disable the orphan_file feature which is not supported by langdale resize2fs.
+# Removing this breaks installation of new RAUC bundles on langdale-based TACs.
```
jluebbe: It's not clear whether `this` refers to the override or the feature. Perhaps:
``` suggestion
# Disable the orphan_file feature which is not supported by langdale's resize2fs and which
# would otherwise break installation of new RAUC bundles on langdale-based TACs.
# The override can be removed when updates from langedale are no longer needed.
```

jluebbe (changes requested):
Looks good otherwise.

meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
hnez: Yeah, that sounds good. I've changed the respective commit and added you suggestion.
I've slightly modified it to limit the line length.

jluebbe (approved):

a3f:
FYI, dt-utils v2023.11.0 incorporates the fix in https://github.com/linux-automation/meta-lxatac/pull/70/commits/62dfa038848123597e994d524f2154d2a7c921e4, so you should be able to drop the patch on your next update.

hnez:
> FYI, dt-utils v2023.11.0 incorporates the fix in [62dfa03](https://github.com/linux-automation/meta-lxatac/commit/62dfa038848123597e994d524f2154d2a7c921e4), so you should be able to drop the patch on your next update.
 
Cool, I've added Pull-Requests in pengutronix/meta-ptx#121 and rauc/meta-rauc#292 to update `dt-utils` there.

PR #71, opened 2023-11-03
meta-lxatac-bsp: emmc-image: clear barebox state on install

hnez:
The bootchooser info stored in the barebox state does not make any sense after installing a fresh image via `fastboot`
(see the removed `README.md` section for more details).

Make sure the barebox state part of the `emmc-image-lxatac.simg` sparse image was touched during image generation so it is not marked as sparse and thus skipped when flashing.

There are also preparations to expose the barebox state as a partition with the correct type `UUID`, but that requires some changes to the barebox devicetree because as of now it tries to add the state twice and fails the second time.

Todo before merging:

- [x] Check if the later partitions (`root-a`, `root-b`, `data`) did not shift around and their alignment is still the same.
  2023-11-10: Partition numbering did indeed change, but this is less of a problem now that we have label-based boot slot selection.
- [x] Check if the barebox state and environment are cleared as intended.
- [x] Remove the workarounds for invalid barebox state from the testsuite

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,24 +500,3 @@ console via the debug UART adapter / the USB console provided by barebox
 and execute `wd -x` on the shell. E.g.:
 
     barebox@Linux Automation Test Automation Controller (TAC):/ wd -x
-
-##### The installed bundle does not start
-
-When flashing the emmc-image only the first of two update partitions is
-initialized. The second one is set up on first boot and filled for the first time
-once you install a RAUC bundle on the system for the first time.
-There is a 50/50 chance that barebox is currently set up to boot from this
-non-existing partition, which will fail and reduce its retry counter.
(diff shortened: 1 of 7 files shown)


hnez:
I've tried to implement two more elegant approaches to what is in this PR, but they did not work out:

- Use `/dev/zero` in the genimage config did not clear the state
- Adding the barebox state partitions to the partition table did not work without breaking backward compatibility.

What _is_ new is that we no longer rely on partition numbers in the boot process and instead use partition labels in barebox. This allows us to change the partition numbering. This PR already uses that to remove the `reserved` partition and a later PR can use it again to add the barebox partitions to the partition table.

I've made sure that the testsuite succeeds and would suggest merging the PR as-is.

jluebbe (approved):

PR #73, opened 2023-11-06
meta-lxatac-bsp: linux-lxatac: upgrade to 6.6

hnez:
umpf tag is now: 6.6/customers/lxa/lxatac/20231106-1 The tag includes the following topic branches:

  - v6.6/customers/lxa/lxatac
  - v6.6/topic/pwm-stm32
  - v6.6/topic/pwm-backlight
  - v6.6/topic/tty-trigger

Starting with this kernel version the LXA TAC base devicetree is mainline \o/ and we no longer have to add it here as a patch.
The TI LMP92064 buffering support also made it mainline.

I've also moved back from fetching the sources via git to tarballs. The benefits are:

- An `uname -a` does not show the kernel version as `-dirty` anymore (caused by the patches we applied to the git repository)
- I get less confused when updating the `PV` does not do anything.
- The initial download is a lot smaller.


hnez:
Assigned @ukleinek because he knows a lot about new `defconfig` flags, making him a good reviewer for this.

meta-lxatac-bsp/recipes-kernel/linux/files/defconfig
```diff
@@ -15,7 +15,7 @@ CONFIG_CC_HAS_ASM_GOTO_OUTPUT=y
 CONFIG_CC_HAS_ASM_GOTO_TIED_OUTPUT=y
 CONFIG_CC_HAS_ASM_INLINE=y
 CONFIG_CC_HAS_NO_PROFILE_FN_ATTR=y
-CONFIG_PAHOLE_VERSION=120
+CONFIG_PAHOLE_VERSION=124
```
ukleinek, line 18: This is an annoying hunk, as it only contains the version of pahole you have installed. IMHO this should not be a kconfig item, but a make runtime variable. (Nothing to blame on you.)

meta-lxatac-bsp/recipes-kernel/linux/files/defconfig
```diff
@@ -2381,7 +2407,9 @@ CONFIG_SPI_STM32_QSPI=y
 CONFIG_SPI_SPIDEV=y
 # CONFIG_SPI_LOOPBACK_TEST is not set
 # CONFIG_SPI_TLE62X0 is not set
-# CONFIG_SPI_SLAVE is not set
+CONFIG_SPI_SLAVE=y
```
ukleinek, line 2410: Is this intended?

ukleinek (commented):
Otherwise looks good to me.

However I'd like to update the topic branches a bit to make the changes a bit easier to understand and give a warmer and still more cosy feeling. I'll prepare another PR to demonstrate that.

hnez:
Closed in favor of #101, which does both the 6.5 -> 6.6 and the 6.6 -> 6.7 update.

PR #74, opened 2023-11-10
meta-lxatac-software: bundles: hook.sh: migrate user-specified files

hnez:
Allow users to migrate custom files between installs.

Users can create files like this:

```bash
$ cat /etc/rauc/migrate.d/git_config.conf
/etc/gitconfig
/home/root/.gitconfig
```

Which will result in the listed files being migrated to the new RAUC slot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,18 @@ bundle on your TAC:
 
 If that does not work you may want to [re-install via USB](#installing-images-via-usb).
 
+If you've made changes to your TACs configuration that you want to preserve,
+which are not covered by the default update migration logic,
+you can add a list of additional files to migrate to the
+`/etc/rauc/migrate.d`:
+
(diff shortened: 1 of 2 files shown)


jluebbe (approved):

jluebbe:
The code looks good. Do we need documentation for this feature?

SmithChart:
> The code looks good. Do we need documentation for this feature?

Yes, documentation is needed. Probably in the `README` here and an in the product documentation.

hnez:
I've added a section to the README (there was no obvious place to put it, so it got a quite prominent spot right at the start now). 
I've also created a PR against the manual in our internal GitLab instance.

Emantor (approved):

jluebbe (approved):

PR #75, opened 2023-11-24
meta-lxatac-bsp: linux-lxatac: upgrade to 6.7

hnez:
umpf tag is now: 6.7/customers/lxa/lxatac/20231124-1. The tag includes the following topic branches:

 - v6.7/customers/lxa/lxatac
 - v6.7/topic/pwm-stm32
 - v6.7/topic/pwm-backlight
 - v6.7/topic/tty-trigger

No interesting changes this time around. The KSZ drive strength HACK got even HACKier because the `#defines` are no longer in scope, so I replaced them with the computed value.

Also right now the build fails because `perf` fails to compile. See the log [log.do_compile.19484.txt](https://github.com/linux-automation/meta-lxatac/files/13460090/log.do_compile.19484.txt) for more information.

TODO before merging:

 - [x] Add the drive-strength configuration to the devicetree. The drive strength configuration code does not quite work as intended yet, but that should change soon and then we can get rid of the HACK.
 - [x] Fix the broken perf build. There is an [oe-core commit](https://github.com/openembedded/openembedded-core/commit/432b4ac20dadb9c08596266fce4f3941d4856c7e) that I copied from.
 - [x] Wait for 6.7 to be released and update this PR to use the released tarball.

Related Pull Requests
---------------

This PR is based on another PRs that should be kept in mind:

- [ ] The `meta-lxatac` pull request that last touched the kernel patch stack: #73


hnez:
Closed in favor of #101, which does both the 6.5 -> 6.6 and the 6.6 -> 6.7 update.

PR #78, opened 2023-11-27
meta-lxatac-software: labgrid-exporter: drop custom service

Bastian-Krause:
Since [1], the only differences between the labgrid-exporter.service here and in meta-labgrid are:

- `Description=`, exporter -> Exporter
- `RequiresMountsFor=`, which is already provided via lxatac-persistent-labgrid-cache as a drop-in
- `Type=`, defaults to simple anyway if `ExecStart=` is specified but neither `Type=` nor `BusName=` are
- `ExecStart=` specifying the Python interpreter explicitly

Since there are no relevant changes left, drop the custom service and let python3-labgrid use the one from meta-labgrid instead.

[1] https://github.com/labgrid-project/meta-labgrid/pull/42


hnez (approved):
Looks good, thanks!
I've tested it to make sure the exporter comes up fine and it does.

PR #79, opened 2023-11-30
dependabot: add configuration to update layer submodules

hnez:
This should result in us getting regular pull requests like these:

dependabot

So we can stay up to date without having to poll the submodules on our own.

It obviously does not free us from having to test the result and review the changes. But one thing at a time.

Thanks to @KarlK90 for the inspiration to do this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "meta-ptx"]
 	path = meta-ptx
 	url = https://github.com/pengutronix/meta-ptx.git
-	branch = langdale
+	branch = master
 [submodule "meta-oe"]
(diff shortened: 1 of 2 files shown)


Emantor (approved):

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "meta-ptx"]
 	path = meta-ptx
 	url = https://github.com/pengutronix/meta-ptx.git
-	branch = langdale
+	branch = master
```
Emantor, line 4: Note that we are going to remove compatibility with `mickledore` in master soon, so we should probably change this to `mickledore`.

PR #80, opened 2023-12-04
Update to Yocto 4.3 - Nanbield

hnez:
This is a draft PR that tracks the nanbield upgrade status.

Todo before merging:

- [x] Check if `android-tools` is sufficiently recent and if not repair the recipe that failed to build.
    I've made sure that the new `fastboot` can flash another LXA TAC using the same commands as before/in the README and it does.
- [x] Make sure NetworkManager builds again.
  This failing was actually a good thing. As it turns out I have used `meta-oe` master together with `poky` nanbield. Now both are used from the nanbield branch.
- [x] Run the testuite.
- [x] Split up the changes into meaningful commits.


hnez:
All of the open tasks in the "Todo before merging" list are now completed 🎉.

I think we are ready to merge this.

hnez:
It turns out we no longer need the `imx-uuu` patch because the issue got fixed upstream. Yay! 🎉.
I love how many local recipes and patches are deleted in this PR.

jluebbe (approved):

PR #81, opened 2023-12-06
meta-lxatac-software: add custom distro config and static uid/gid config

hnez:
By default user and group IDs are assigned based on recipe build order. This works as long as no files are shared between different builds. In our case we have two cases where files are shared between builds:

- Files migrated via RAUC hooks between rootfs A and rootfs B.
- Files in the `/srv` data partition.

Files in one of these categories that are not owned by root may have the wrong owner uid and gid based on the order recipes were built in for the respective bundle.

Fix that by assigning static user ids and group ids. The `passwd` and `group` files are based on the files found in the 23.11 stable release bundles and images found on our `downloads.linux-automation.com` server, giving us good backward/forward compatibility.
The configuration is rather straight forward and follows the yocto manual entry about [`USERADDEXTENSION`](https://docs.yoctoproject.org/ref-manual/variables.html#term-USERADDEXTENSION).

The uid/gid config is neither a machine-specifc or local configuration and instead belongs into a distro config. Up until now we just used the [`ptx`](https://github.com/pengutronix/meta-ptx/blob/master/conf/distro/ptx.conf) distro from meta-ptx.

Since the vanilla 🍦 `ptx` distro no longer worked for us I decided to make `tacos` instead 🌮, I hope you like it.


meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -0,0 +1,50 @@
+DISTRO = "tacos"
+DISTRO_NAME = "TAC OS - The LXA TAC operating system"
+DISTRO_VERSION = "23.11+dev"
+DISTRO_CODENAME = "tacos-mickledore"
```
hnez, line 4: The convention for `DISTRO_VERSION` should be to set it to the anticipated release version once we start feature freezing and to `+dev` once we released and start to add features again.

The convention for `DISTRO_CODENAME` should be to update it whenever we change yocto releases.

meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -0,0 +1,50 @@
...
+
+LOCALCONF_VERSION = "2"
+
+DISTRO_FEATURES = "argp ext2 largefile usbgadget usbhost wifi xattr nfs zeroconf multiarch systemd usrmerge"
+
+# Select systemd as init manager
+INIT_MANAGER ?= "systemd"
+
+DISTRO_EXTRA_RDEPENDS += "packagegroup-core-boot ptx-profile"
+DISTRO_EXTRA_RRECOMMENDS += "kernel-module-af-packet"
```
ejoerns: These two lines could be omitted I guess. They were copied from poky.conf to ptx.conf once but are not really relevant to the system (or explicitly added anyway).

ejoerns (approved):
Yummy! :taco:

meta-lxatac-software/conf/distro/tacos.conf
hnez: I've removed the two lines now. Starting with as little cargo-culting as possible sounds good to me.

Emantor (approved):

meta-lxatac-software/conf/distro/tacos.conf
jluebbe: Perhaps add a comment here to avoid mistakes in the future?

PR #82, opened 2023-12-07
README: enable an update channel after install from image

hnez:
The releases in the stable and testing channel on `downloads.linux-automation.com` are byte-for-byte equivalent, except for the RAUC bundles, which are re-signed before they are uploaded.

When installing a testing or stable bundle the update channel to enable in the new slot is chosen based on the certificate the bundle was signed with.

For a from-scratch installation we do not have this information, so the user has to choose a channel to enable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,19 @@ Next you can upload the required pieces of software:
 You are now done flashing the LXA TAC and can remove the pull-down on `BT1`
 and power cycle it to boot into your newly flashed image.
 
+### Enabling an update channel
+
+The official filesystem images do not enable an update channel by default.
+This means you will not be notified about updates and can not install official
+RAUC bundles from `downloads.linux-automation.com`.
+
+Use `rauc-enable-cert` to enable the stable update channel and
+`rauc-disable-cert` to prevent RAUC bundles signed with the insecure devel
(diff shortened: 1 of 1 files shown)


jluebbe (changes requested):
We discussed this internally: we think the devel cert should not be enabled by default in manually installed images.

SmithChart:
This means that we have to enable one channel on production devices for sale during bringup, right?

jluebbe:
> This means that we have to enable one channel on production devices for sale during bringup, right?

Or it is enabled during setup with the wizard.

hnez:
> > This means that we have to enable one channel on production devices for sale during bringup, right?
> Or it is enabled during setup with the wizard.

We do not currently enable/disable update channels from within the tacd. That would be an interesting feature and would also be required to allow initial deployment of new-out-of-the-box TACs via RAUC bundle, but that has not been a priority yet.

hnez:
> This means that we have to enable one channel on production devices for sale during bringup, right?

Yes. It looks like we likely messed that up for our first batch of external beta-tester TACs, but we've since added such a step to the bringup tooling.

hnez:
I'll close this PR in favor of #83.

PR #85, opened 2023-12-12
Fix optional shellcheck warnings for hook.sh

Emantor:
Fix some minor optional shellcheck warnings within hook.sh.


hnez (approved):
Very nice!

This PR inspired me to add a CI job for shellcheck to meta-lxatac. See #86 for that.
We have to figure out why it does not run though, but I can already tell that there is a lot more potential for improvement.

But as an incremental improvement this is already great.

PR #87, opened 2023-12-15
CI: add distribution version consistency checks

hnez:
This should make sure that we do not forget to update the `DISTRO_VERSION` and `DISTRO_CODENAME` when a new tag is created or a new yocto version branch is started.

This is an alternative to the proposed comment from a comment in https://github.com/linux-automation/meta-lxatac/pull/81#discussion_r1424148016 because comments are not resistant to being ignored.
A failing CI job should be harder to ignore.

### DIFF
--- a/.github/workflows/distribution-version.py
+++ b/.github/workflows/distribution-version.py
@@ -1,0 +1,91 @@
+import os
+import re
+
+import bb.tinfoil
+import git
+
+# TODO: add minor version numbers once we have them
+RE_VERSION_TAG = "v(?P<year>[0-9][0-9])\.(?P<month>[0-9][0-9])"
+RE_VERSION_BB = "(?P<year>[0-9][0-9])\.(?P<month>[0-9][0-9])(?P<dev>$|\+dev)"
+
+
+def bb_variables(names):
+    ret = dict()
+
+    with bb.tinfoil.Tinfoil(tracking=True, setup_logging=True) as tinfoil:
+        tinfoil.prepare(quiet=2, config_only=True)
+        d = tinfoil.config_data
+
+        for name in names:
+            ret[name] = d.getVar(name)
+
+    return ret
+
+
+def git_prev_tag():
+    repo = git.Repo()
+
+    # e.g. 'v23.11'
+    tag = repo.git.describe(tags=True, abbrev=0)
+
+    # e.g. 'v23.11-40-ga4a1'
+    commit_description = repo.git.describe(tags=True)
+
+    # The description matches the tag if there are no commits since the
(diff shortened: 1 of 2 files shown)


.github/workflows/distribution-version.py
```diff
@@ -0,0 +1,89 @@
...
+    repo = git.Repo()
+
+    current_commit_is_tagged = True
+
+    for commit in repo.iter_commits():
+        for tag in repo.tags:
+            if tag.commit == commit:
+                return (tag, current_commit_is_tagged)
+
+        current_commit_is_tagged = False
```
jluebbe: Perhaps use something based on `repo.git.describe(tags=True, abbrev=0)`?

.github/workflows/distribution-version.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if commit_is_tagged:
+        # The version in a tagged commit must match the version in the tag's name.
+        assert version_bb["dev"] == ""
+        assert version_tag_numeric == version_bb_numeric
+
+    elif version_bb["dev"] == "":
+        # Release candidates already have the next release version set,
+        # but there must not be a tag with the same or a newer version number
+        # in the current commits history.
```
jluebbe:
```suggestion
        # Release candidates already have the next release version set,
        # but it must be newer than any tag in the current commit's history.
```

.github/workflows/distribution-version.yaml
```diff
@@ -0,0 +1,23 @@
+name: distribution version
+
+on:
+  push:
+  pull_request:
```
jluebbe, line 5: Also on tags?

jluebbe (changes requested):

.github/workflows/distribution-version.py
hnez: Oh cool, I did not find anything regarding `describe` directly in `repo` so I hand crafted this "`describe`".
Did not dig deep enough into the documentation to find `repo.git.…`.
I've updated the PR now.

.github/workflows/distribution-version.py
hnez: Yep, that's a less round-about description. I've copied the recommendation into my updated PR.

.github/workflows/distribution-version.yaml
hnez: Makes sense. A quick glimpse at the documentation tells me that [`on: create`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#create) is what I want?

jluebbe (approved):

PR #88, opened 2023-12-15
meta-lxatac-software: tacd: failsafe: gpiofind no longer exists

hnez:
The gpiofind command was removed, so the tacd-failsafe command would currently fail (which is kind of ironic).

Instead gpioset can be directly supplied with the GPIO names. The GPIOs are not guaranteed to keep the assigned value once gpioset exits, so we just keep it running for some time before trying to restart the tacd.

### DIFF
--- a/meta-lxatac-software/recipes-rust/tacd/files/tacd-failsafe.sh
+++ b/meta-lxatac-software/recipes-rust/tacd/files/tacd-failsafe.sh
@@ -1,7 +1,17 @@
 #!/bin/sh
 
 # Disable DUT power and enable discharge resistor (both lines are active low).
-# gpiofind returns a chip + id pair in the form of: gpiochip0 2.
-# We can not use quoting here as the command should be: gpioset gpiochip0 2=1.
-gpioset $(gpiofind DUT_PWR_EN)=1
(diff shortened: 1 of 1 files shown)


meta-lxatac-software/recipes-rust/tacd/files/tacd-failsafe.sh
```diff
@@ -3,5 +3,6 @@
 # Disable DUT power and enable discharge resistor (both lines are active low).
 # gpiofind returns a chip + id pair in the form of: gpiochip0 2.
```
jluebbe: Please update the comment as well.

Also describe why it's useful to keep the gpioset running.

jluebbe (changes requested):

meta-lxatac-software/recipes-rust/tacd/files/tacd-failsafe.sh
hnez: Heh. I completely missed the comment when I changed the code.
I've now written a (bit lengthy) comment about the hows and whys of `tacd-failsafe.sh`.

jluebbe (approved):

PR #91, opened 2023-12-20
meta-lxatac-bsp: libtraceevent: backport "build with Meson" from oe-core

hnez:
This fixes an issue where the libtraceevent build would sometimes generate .so files with some objects missing due to a [race condition in the Makefile](https://git.yoctoproject.org/poky/commit/?id=142b6921bde5b22bcd7e0e5e8d895575d0fb6c64).
One of these builds made it into the sstate of our CI builds, making them fail.

The upstream fix is to just move to meson as a build system, so lets just do the same.

This only affects our mickledore branch, as the backport comes from the nanbield poky branch (and langdale did not use libtraceevent).

Thanks to @ejoerns for finding the respective oe-core patch!

### DIFF
--- a/meta-lxatac-bsp/recipes-backport/libtraceevent/libtraceevent/meson.patch
+++ b/meta-lxatac-bsp/recipes-backport/libtraceevent/libtraceevent/meson.patch
@@ -1,0 +1,74 @@
+Fixes for the Meson build of libtraceevent:
+
+- Make the plugin directory the same as the Makefiles
+- Install the plugins as modules not static and versioned shared libraries
+- Add an option to disable building the documentation (needs asciidoc and xmlto)
+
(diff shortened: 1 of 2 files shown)


hnez:
It would be great if we could get this merged. I would like to leave the mickledore branch behind soon - but not in a broken state.

ejoerns:
@hnez I would put it in `recipes-backport/` since this is not only making it build with meson but also includes the version bump to `1.7.3`. (And libtraceevent moved from `meta-oe` to `oe-core` between `mickledore` and `nanbield`). Otherwise fine for me.

hnez:
Makes sense. I've moved the recipe to `recipes-backport`.

ejoerns (approved):

PR #93, opened 2023-12-20
meta-lxatac-software: bcu: add recipe for nxp board control utilities

hnez:
The `bcu` command is used to communicate with board control hardware on NXP development boards to:

  - reset / power cycle the board into a specified boot mode
  - monitor the power consumption
  - eeprom read / write
  - temperature measurements

Sadly [the board I tried to use this with](https://www.nxp.com/design/design-center/development-boards/i-mx-evaluation-and-development-boards/i-mx-8quadxplus-multisensory-enablement-kit-mek:MCIMX8QXP-CPU) is not actually supported, so I could only check that it builds and shows the help message.
It also discovered the FTDI device on the board, so that seems to work as well.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/bcu/bcu/0001-CMakeLists-do-not-use-vendored-libcurl.patch
+++ b/meta-lxatac-software/recipes-devtools/bcu/bcu/0001-CMakeLists-do-not-use-vendored-libcurl.patch
@@ -1,0 +1,38 @@
+From ef1aaeee984ab13dab5d46c73e9b207cbdf62c55 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Leonard=20G=C3=B6hrs?= <l.goehrs@pengutronix.de>
+Date: Wed, 20 Dec 2023 14:37:20 +0100
+Subject: [PATCH] CMakeLists: do not use vendored libcurl
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Upstream-Status: Inappropriate [upstream ticket https://github.com/nxp-imx/bcu/issues/17]
+
+Signed-off-by: Leonard Göhrs <l.goehrs@pengutronix.de>
+---
+ CMakeLists.txt | 5 +++--
+ 1 file changed, 3 insertions(+), 2 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 3475179..bfecdb0 100644
+--- a/CMakeLists.txt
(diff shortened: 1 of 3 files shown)


jluebbe:
There is a open PR (https://github.com/labgrid-project/labgrid/pull/1273) to support BCU via labgrid.

jluebbe:
What is the status of the libcurl patch? Could it be upstreamed?

hnez:
> What is the status of the libcurl patch? Could it be upstreamed?

I had a quick look at it and it looks to me like that would require changes to the build system for macOS, linux and windows, making testing the changes a chore.
Since we do not have a real usecase for `bcu` yet and only wanted to check it out for fun, I don't think it is worth it to put in the effort.

I'd say we close this PR for now and have a second look once an actual usecase arises.

jluebbe:
> > What is the status of the libcurl patch? Could it be upstreamed?
> 
> I had a quick look at it and it looks to me like that would require changes to the build system for macOS, linux and windows, making testing the changes a chore. Since we do not have a real usecase for `bcu` yet and only wanted to check it out for fun, I don't think it is worth it to put in the effort.
> 
> I'd say we close this PR for now and have a second look once an actual usecase arises.

I'd also be OK with at least reporting this as an issue to upstream and then pointing to that issue in a patch status header. Perhaps that way, it's solved upstream by the time we get around to updating it.

meta-lxatac-software/recipes-devtools/bcu/bcu/0001-CMakeLists-do-not-use-vendored-libcurl.patch
```diff
@@ -0,0 +1,38 @@
+From ef1aaeee984ab13dab5d46c73e9b207cbdf62c55 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Leonard=20G=C3=B6hrs?= <l.goehrs@pengutronix.de>
+Date: Wed, 20 Dec 2023 14:37:20 +0100
+Subject: [PATCH] CMakeLists: do not use vendored libcurl
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Upstream-Status: Inappropriate [upstream ticket https://github.com/nxp-imx/bcu/issues/17]
```
hnez, line 9: I've opened an issue in the bcu repository nxp-imx/bcu#17 and added an Upstream-Status line to the patch.

hnez:
Hi @jluebbe,

now that we have an `Upstream-Status` entry, do we want this recipe in meta-lxatac?

jluebbe (approved):

PR #95, opened 2024-01-04
(mickledore) meta-lxatac-software: passwd: reserve uid/gid for systemd-journal-upload

hnez:
A backport of #92 for mickledore. See the description there for details.


hnez:
I've made sure this builds. Since #92 is already reviewed I think this is enough to merge this.

PR #96, opened 2024-01-11

hnez:
This brings a couple of changes:

 - The RGB status LED is now dimmed a bit on idle via PWM. This makes all LEDs the same brightness on idle. When e.g. flashing the locator the status LED can however be made brighter than the other LEDs.
 - The Display now shows reboot notifications when the other RAUC slot would be booted on reboot. (Previously this check was based on which slot contained the newer image - even if the newer slot would not be booted after a suggested reboot).
 - The setup wizard was simplified. The "setup via rauc bundle" setup mode (that was not implemented yet) was removed. There is now only one setup method.
 - The tacd now asks for consent before polling for updates from the update server.
 - The misleading "Your TAC is overheating" warning if the connection between the web interface and tacd server breaks should be gone.

### DIFF
--- a/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
+++ b/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
@@ -3,7 +3,7 @@ inherit cargo
 DEFAULT_PREFERENCE = "-1"
 
 SRC_URI += "git://github.com/linux-automation/tacd.git;protocol=https;branch=main"
-SRCREV = "803b2084b26634542ebf763c818083884d49ac0d"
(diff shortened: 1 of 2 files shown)


hnez:
Hi @jluebbe,

I think merging this should be okay, because reviewing `tacd` happens in the respective repository and this change only updates the git commit that is used.
What do you think?

hnez:
Hi @jluebbe,

I think it would be great if we could get this merged.

jluebbe (approved):

jluebbe:
> I think merging this should be okay, because reviewing `tacd` happens in the respective repository and this change only updates the git commit that is used. What do you think?

I've not looked at the tacd commits in the context of this PR.

PR #98, opened 2024-01-19
meta-lxatac-bsp: rauc: add patch to fix RAUC info / update notifications

hnez:
Currently the update notifications on the LXA TAC LCD and web interface do not work because the tacd can not get informations about the RAUC bundle on the update server via RAUC's DBus API.

This is because the process that RAUC spawns to handle the nbd bridge for http streaming crashes due to an unitialized pointer access.

This patch is taken from RAUC pull request rauc/rauc#1332 , which fixes the issue.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/rauc/files/0001-src-nbd-initialize-info_headers-to-NULL.patch
+++ b/meta-lxatac-bsp/recipes-core/rauc/files/0001-src-nbd-initialize-info_headers-to-NULL.patch
@@ -1,0 +1,38 @@
+From 967d32571e226d15c345b2d8cec932a651716f4b Mon Sep 17 00:00:00 2001
+From: Enrico Jorns <ejo@pengutronix.de>
+Date: Thu, 18 Jan 2024 16:49:46 +0100
+Subject: [PATCH] src/nbd: initialize info_headers to NULL
+
+When g_variant_dict_lookup() does not find the given key, it does not
+modify the values of the arguments provided.
+
+This results in info_headers being undefined and can cause the
+rauc-nbd-server to crash, which will result in a broken request on
+client side that lets RAUC terminate hard with:
+
+(rauc:769): rauc-nbd-ERROR **: 15:40:30.553: failed to recv nbd config reply header
+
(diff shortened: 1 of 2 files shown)


meta-lxatac-bsp/recipes-core/rauc/files/0001-src-nbd-initialize-info_headers-to-NULL.patch
```diff
@@ -0,0 +1,36 @@
...
+
+This results in info_headers being undefined and can cause the
+rauc-nbd-server to crash, which will result in a broken request on
+client side that lets RAUC terminate hard with:
+
+(rauc:769): rauc-nbd-ERROR **: 15:40:30.553: failed to recv nbd config reply header
+
+The behavior of g_variant_dict_lookup() has also recently been clarified
+in the glib documentation:
+https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3334
```
ejoerns, line 17: You might want to add an [Upstream-Status](https://docs.yoctoproject.org/contributor-guide/recipe-style-guide.html#patch-upstream-status) like:

    Upstream-Status: Submitted [https://github.com/rauc/rauc/pull/1332]

hnez:
> You might want to add an Upstream-Status like:

Sounds good. Is this the correct place to put it:

https://github.com/linux-automation/meta-lxatac/compare/f24eb46b0ac2732da3b708d66c8e07abc87dc08e..8eb01d4382db1ac543823dfb7dba4141f429bc4e

?

ejoerns:
> > You might want to add an Upstream-Status like:
> 
> Sounds good. Is this the correct place to put it:
> 
> https://github.com/linux-automation/meta-lxatac/compare/f24eb46b0ac2732da3b708d66c8e07abc87dc08e..8eb01d4382db1ac543823dfb7dba4141f429bc4e
> 
> ?

I do think so, yes. :+1:

hnez:
Cool! Then I think we should get this merged sooner than later to prevent users from gettting stuck on LXA TAC images without update notifications.

ejoerns (approved):

PR #100, opened 2024-01-25
README: Add contribution guidelines with DCO

SmithChart:
We want to help new contributors help find their way to the 'Pull Request' tab.

Also we want contributions to contain a DCO.

PR #101, opened 2024-01-25
Kernel update to 6.7

ukleinek:
This is heavily inspired by @hnez's PR https://github.com/linux-automation/meta-lxatac/pull/73 . The resulting image should be quite similar, the upside is mostly better documented and more uptodate patches.

For now this is only build tested.


hnez:
I'am about to runtime test this now. Could you add `Signed-off-by:`s based on the DCO in #100 in the meantime?

hnez:
I've tested the patch now and ran the CI testsuite. Everything seems to be working! 🎉

Now we just need to get #100 merged and S-o-bs added to the commits, then this can be merged.

A bit pedantic, but when you have to touch the commit messages anyways, could you modify the `linux-lxatac: Version bump to 6.6 ` commit message a bit? It contains the following:

```
…
The only other relevant change is that the RGMII clock line's drive
strength is reduced using device tree properties. See
0003-ARM-dts-stm32-lxa-tac-reduce-RGMII-interface-drive-s.patch
Note however this only starts working once commit v6.7-rc1~160^2~325^2
("net: dsa: microchip: Add drive strength configuration") is added to
the stack.
```

which implies that the mentioned commit does what we need and removes the need for the drive strength hack, which is not the case, because the drive strength setting in said commit is not re-applied after a reset of the switch (of which there are multiple in the drivers init sequence).

hnez:
Hi @ukleinek,

PR #100 is now merged, so we now have a defined DCO policy. Are you okay with the policy and adding your S-o-b now?

ukleinek:
Rebased on top of nanbield and added the now correctly formalized S-o-b lines

hnez:
I ran a quick build test to make sure nothing broke with the meta layer updates that were merged in between.

Looks like nothing broke, so we are ready to merge. Thanks @ukleinek!

hnez (approved):

marckleinebudde:
Can we close this PR?

PR #103, opened 2024-02-06
meta-lxatac-software: remove upstreamed recipes

hnez:
We have accumulated a few recipes that are no longer required due to them being updated upstream (`uhubctl`) or upstreamed (`linux-serial-test` and `smemstat` (thanks to @jluebbe for upstreaming both of them)).

Remove them to simplify meta-lxatac and also to benefit from upstream updates (both `linux-serial-test` and `smemstat` in meta-oe are newer than what we have in meta-lxatac).


hnez:
Closed in favor of #108, which does the same.

PR #104, opened 2024-02-08
linux-lxatac: Drop unneeded pwm-bl patch

ukleinek:
Since the pwm hardware driver (here: pwm-stm32) has .get_state() implemented the extra effort to keep the pwm-bl state on probe isn't needed any more.

This brings us one patch nearer to the goal of using Linux without patches.


hnez:
I've tested the same change before and back then it did not result in renewed backlight flicker, so I think this should be fine.

One patch closer to mainline, thanks!

hnez (approved):

PR #105, opened 2024-02-09
meta-lxatac-bsp: rauc: split rauc.bb and rauc-conf.bb

hnez:
This PR is made in anticipation of a change in rauc/meta-rauc#147, which splits the rauc recipe into a recipe for the binaries and one for the configuration.

TODO before merging

- [x] Wait for the change to land in `meta-rauc` and change the submodule URL back to the upstream project.


meta-lxatac-bsp/recipes-core/rauc/rauc-conf.bbappend
```diff
@@ -0,0 +1,52 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+DEPENDS:append = "openssl-native"
+RDEPENDS:${PN}:append = "bash"
```
ejoerns: As long as no overrides are used, I'd use simple `+=` for `DEPENDS` and `RDEPENDS` since they are easier to parse and this is consistent with other recipes. However, for the scope of this PR I'd leave it as it was before the split-up, anyway. (Same for `SRC_URI` below.)

ejoerns (approved):

meta-lxatac-bsp/recipes-core/rauc/rauc_%.bbappend
```diff
@@ -1,8 +1,8 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-RDEPENDS:${PN}:append = "e2fsprogs-resize2fs"
+RDEPENDS:${PN} += " e2fsprogs-resize2fs"
```
ejoerns: Spaces at the beginning of a string are not required when using "+=".
In fact, they were required before when using `:append` :wink:

hnez: \*sigh\*. Fixed

ejoerns (approved):

hnez:
I've just checked if this PR builds and that `rauc info` and update notifications work with rauc 1.11.2.
Look good, thanks for the release and review!

PR #109, opened 2024-02-18
Clean Up of lxatac Machine Config

ejoerns:
Cleans up and moves configurations that should be set in other locations.


hnez:
Hi,

I think the commit message for the second commit was the wrong way around:

    DISTRO_FEATURES should be defined in the machine, rather than in the distro.

I've turned it around to:

    DISTRO_FEATURES should be defined in the distro, rather than in the machine.

hnez (approved):
With the commit message changes everything looks fine to me.

PR #110, opened 2024-02-18
Various Minor Recipe Cleanups

ejoerns:
* Removes ineffective and commented-out lines
* adds missing SUMMARY to recipes
* runs oe-stylize to have a more standard structure


hnez (approved):
Hi,

great pull request, I've learned a lot about best practices and bitbake in general by just reading through the commits.

It would be great to automate some of the recipe file style checks. I've created a draft PR that runs `oe-stylize.py` as a GitHub action (#114). As you have already commented there this may not work out due to false positives, but maybe we could get it to work. I think that would be cool.

But as for _this_ PR: LGTM, thanks!

PR #113, opened 2024-02-18
lxatac-core-image-base: Apply Some Cleanups

ejoerns:
This is a first step into reducing the quite long and very explicit `IMAGE_INSTALL` list by oe-core default mechanisms like `IMAGE_FEATURES` and feature-controlled packagegroups.

### DIFF
--- a/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
+++ b/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
@@ -65,7 +65,6 @@ IMAGE_INSTALL:append = "\
     iproute2-rtacct \
     iproute2-ss \
     iproute2-tc \
-    iw \
     libdrm-tests \
     libgpiod-tools \
     libiio \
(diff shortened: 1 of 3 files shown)


meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
```diff
@@ -3,6 +3,7 @@ SUMMARY = "LXA TAC image containing a selection of useful development tools"
 BAD_RECOMMENDATIONS = "rng-tools"
 
 IMAGE_FEATURES = "ssh-server-openssh empty-root-password"
+IMAGE_FEATURES += "tools-debug"
```
hnez: I think either all of the `IMAGE_FEATURES` should be assigned at the same time or all should be `+=`-ed, but not half and half.

hnez (approved):
I think these changes are fine, but would prefer a second pair of eyes.
@jluebbe what do you think?

hnez:
@ejoerns please rebase to solve the conflict in `tacos.conf`.

meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
ejoerns: Agreed and fixed.

jluebbe (approved):

PR #114, opened 2024-02-19
CI: add a linting step for all of our custom bitbake recipes

hnez:
This PR is inspired by #110, which mentions running meta-openembedded's [`oe-stylize.py`](https://github.com/openembedded/meta-openembedded/blob/master/contrib/oe-stylize.py) on the bitbakes files in `meta-lxatac`.

Now that I have learned that there is a linter script for bitbake files I want it as a GitHub action.

So here is a pull request to add it.

TODO before merging:

  - [ ] Make sure the action runs without errors

### DIFF
--- a/.github/workflows/oe-stylize.sh
+++ b/.github/workflows/oe-stylize.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e -u -o pipefail
+
+FILES="$(find meta-lxatac-bsp/ meta-lxatac-software/ -type f -name \*.bb)"
+
+result="0"
+
+while read -r recipe
+do
+    # The oe-stylize script is not parameterizable, so we can not tell it not
+    # to output warnings about variables it does not know.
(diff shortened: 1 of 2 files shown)


ejoerns:
@hnez While I like the idea in general, I have some objections against running this in CI: The script is not too actively maintained (primarily got some PEP8-related fixes over the last years) and is not part of oe-core. I have seen several false positives I had to re-fix manually in some recipes (like when properly using self-defined variables).

Thus I'd say it is a good tool for running manually on your recipe, but I am not sure if we should enforce it in CI automatically.

hnez:
I've just rebased the PR, just to have a second look at the output of the job.

On the one hand I really like having one canonical version of source files with their formatting defined by a formatter, on the other hand `oe-stylize.py` does not seem to be up to the task yet, like `rustfmt`, `prettier` or `clang-format` are.

I will close the PR for now. Maybe some day we will have an appropriate auto-formatter.

PR #115, opened 2024-02-23
meta-lxatac-bsp: lxatac-factory-data: order service before udev

hnez:
This is an altternative to #107. Since both approaches are valid and have their benefits and drawbacks I made this a competing PR.

Order the factory data collecting service before the udev systemd service to make sure that the `.link` file it creates exists before udev starts looking for rules.

This should fix an issue we have that results in the `tac-bridge` interface having the wrong MAC address on some boots.

In contrast to #107 it does so without triggering udev to reload its rules.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.service
+++ b/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.service
@@ -1,10 +1,12 @@
 [Unit]
 Description=Gather and Distribute Factory Data
-Before=NetworkManager.service
+DefaultDependencies=no
+After=systemd-remount-fs.service
+Before=systemd-udevd.service
 
 [Service]
 Type=oneshot
 ExecStart=/usr/sbin/lxatac-factory-data
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=sysinit.target
--- a/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
+++ b/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
@@ -67,4 +67,3 @@ OriginalName=tac-bridge
 MACAddress=${MAC_BRIDGE}
 MACAddressPolicy=none
 EOF
-
--- a/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
+++ b/meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
@@ -11,14 +11,14 @@ DST_LINK_FILE_BASE="/run/systemd/network/"
 mkdir -p "${DST_ENVS_BASE}"
 mkdir -p "${DST_LINK_FILE_BASE}"
 
-# Set the static hostname if it is not set yet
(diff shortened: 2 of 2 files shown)


meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.service
```diff
@@ -1,10 +1,13 @@
 [Unit]
 Description=Gather and Distribute Factory Data
-Before=NetworkManager.service
+DefaultDependencies=no
+After=systemd-remount-fs.service
+Before=systemd-udevd.service
+RequiresMountsFor=/run
```
michaelolbrich: `/run` is always mounted before any service starts, so this line is unnecessary.

meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
```diff
@@ -67,4 +65,3 @@ OriginalName=tac-bridge
 MACAddress=${MAC_BRIDGE}
 MACAddressPolicy=none
 EOF
-
```
michaelolbrich, line 70: Unrelated change.

meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
```diff
@@ -11,14 +11,12 @@ DST_LINK_FILE_BASE="/run/systemd/network/"
...
-# (The transient hostname is passed via systemd.hostname= commandline)
-# --------------------------------------------------------------------
+# If no hostname is set yet persist the one we currently have
+# -----------------------------------------------------------
 
 if [[ ! -e /etc/hostname ]]
 then
-    hostname="$(hostnamectl --transient hostname)"
-    hostnamectl --static hostname "${hostname}"
+    hostname > /etc/hostname
```
michaelolbrich, line 21: This change deserves to be mentioned in the commit message.

michaelolbrich (changes requested):

meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.service
hnez: I did a quick `rg RequiresMountsFor` before adding that to check if other service files in the wild had it and there were some.
But if we do not need it that's even better.

Dropped the `RequiresMountsFor=`.

meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
hnez: Split into a separate commit.

meta-lxatac-bsp/recipes-core/lxatac-factory-data/files/lxatac-factory-data.sh
hnez: Split into a separate commit with some info in the commit message.
Also added more information in the comment.

michaelolbrich (approved):
The correct spelling is "D-Bus" but in the commit message it does not really matter.
The rest looks good

PR #117, opened 2024-03-01
poky/meta-*: update to most recent nanbield branches

hnez:
Just a regular maintenance update.

The most interesting change is that we no longer need the workaround with the zeroed-out images to get genimage to fill our state partitions with zeros (instead of leaving them sparse).


ejoerns (approved):
LGTM.

Not sure if switching to a defined stable release version for poky could be even more appropriate / stable, but using the latest 'nanbield' branch state should be fine as well.

PR #121, opened 2024-03-13

### LINKED ISSUE #118: Update tacd the "second release" version
Once the milestone https://github.com/linux-automation/tacd/milestone/1 there is resolved we need to bump the version in the `tacd` and `tacd-webinterface` recipes.

hnez:
This brings a couple of changes:

Web:

  - Labgrid configuration files are now rendered as text in the web interface to prevent syntax error warnings cause by the jinja templating inside these files.
  - Better inform users about which data is sent when polling for updates is enabled.

Daemon:

  - Only check for DUT power supply errors when the DUT power supply is on. This should solve errors where the TAC refuses to turn the power supply on when a noisy power supply is connected.
  - The on display on the device now shows a legend of which button does what and improve the help screen section about which button does what.
  - No longer fail when unknown keys are found tacds state file, e.g. after a downgrade.

TODO before merging:

  - [x] Test an image with the new tacd on actual hardware
        20240313 lgo: assigned to me until that is done.

This fixes #118.

### DIFF
--- a/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
+++ b/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
@@ -3,7 +3,7 @@ inherit cargo
 DEFAULT_PREFERENCE = "-1"
 
 SRC_URI += "git://github.com/linux-automation/tacd.git;protocol=https;branch=main"
-SRCREV = "e94c996e95fc215ef86deb1dd130205582eafeaa"
+SRCREV = "740ffaca3eb8560a556f7d942083928f290d8a19"
 S = "${WORKDIR}/git"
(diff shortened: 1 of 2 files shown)


SmithChart:
Nitpick:

In the commit message:
```
  - The on display on the device now shows a legend of which button
    does what and improve the help screen section about which button
    does what.
```
Sounds like there's one "on" too many there.

hnez:
> Sounds like there's one "on" too many there.

Oh yes, yes indeed. Fixed! Thanks

hnez:
I've tested the new tacd on hardware and it looks really good (if I may say so myself).

I think we should get this merged and roll a new release.

SmithChart (approved):
LGTM

PR #122, opened 2024-03-19
Apply review results from a meta-lxatac derived BSP

hnez:
I've based a (currently non-public) BSP on meta-lxatac, which was in turn reviewed by @Bastian-Krause, who provided great input on how to improve said BSP.

Since the BSP is based on meta-lxatac, some of these review results also apply here, so I am going through them and applying them here as well.

TODO before merging:

  - [x] Go through the complete review and apply the findings.
  - [x] Decide how to provide proper attribution to @Bastian-Krause, by e.g. making him the git author or adding a note to the commits.
  - [x] Test the changes to make sure they do not accidentally break something.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/lxatac-repart/lxatac-repart.bb
+++ b/meta-lxatac-bsp/recipes-core/lxatac-repart/lxatac-repart.bb
@@ -1,3 +1,4 @@
+SUMMARY = "Use systemd's update mechanism to only re-partition the disk after image install"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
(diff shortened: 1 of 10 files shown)


meta-lxatac-bsp/recipes-core/lxatac-repart/lxatac-repart.bb
```diff
@@ -1,3 +1,4 @@
+SUMMARY = "Use systemds update mechanism to only re-partition the disk after image install"
```
Bastian-Krause:
```suggestion
SUMMARY = "Use systemd's update mechanism to only re-partition the disk after image install"
```

Bastian-Krause (approved):

meta-lxatac-bsp/recipes-core/lxatac-repart/lxatac-repart.bb
hnez: Sounds good. I've changed the SUMMARY to match your suggestion.

PR #125, opened 2024-03-22
meta-lxatac-bsp: barebox: update to 2024.03.0 with Gen 3 support 

hnez:
This PR updates the barebox bootloader version on the TAC to version 2024.03.0 and adds support for the most recent hardware generation (Gen 3).

### Related Pull Requests

This PR is based on another PRs that should be merged first:

- [x] The `meta-lxatac` pull request that last touched the barebox patch stack: #94


hnez:
Hi @a3f, would you be interested in reviewing this pull request and also have a look at the ones leading up to this one?
We would like to release a new meta-lxatac / tacos stable release containing LXA TAC hardware generation 3 support some time this week to get ready to ship devices, so it would be great to get this reviewed sooner than later.

README.md
```diff
@@ -486,6 +486,11 @@ Next you can upload the required pieces of software:
     # Nothing is being flashed permanently yet.
     $ dfu-util --alt 3 -D emmc-boot-image-lxatac.fip
 
+    # Start a fastboot command in the background that disables the watchdog
+    # once the LXA TAC shows up as fastboot device.
+    # This also stops the automatic boot process.
+    $ fastboot oem exec "wd -x" &
```
a3f, line 492: AFAIK, the watchdog is only enabled before booting Linux (`boot` command), so `wd -x` directly after DFU bring-up doesn't matter for watchdogs. As for stopping the automatic boot process, `fastboot flash` has the same effect, see:
https://github.com/barebox/barebox/blob/5423885658da4beddec619c1ec4b1cb741c68554/common/fastboot.c#L814

If this change fixes something, I suspect the issue is elsewhere.

a3f: What I usually do is have an init script:
```
#!/bin/sh

if [ "$bootsource" = serial ]; then
   global.autoboot_timeout=30
fi
```

meta-lxatac-bsp/recipes-bsp/barebox/barebox_2023.10.0.bb
```diff
@@ -6,10 +6,12 @@ SRC_URI += " \
 "
 require files/patches/series.inc
 
-SRC_URI[sha256sum] = "d0f78a69ba240327247c8fd0e1d45287e4a0dff99ed847e9a696cc2da0cf388c"
+SRC_URI[sha256sum] = "07823aa3cdc083087788a1cccded99a15a250746e259c2ff905dc48f8d9f409a"
 
 COMPATIBLE_MACHINE = "lxatac"
 
+BAREBOX_IMAGES = "*.stm32"
```
a3f: It would be nice if we could import the [barebox recipe that eventually should make it into OE-core](https://lore.kernel.org/all/20230425184720.456896-1-ejo@pengutronix.de/) in a future update. This parses the `barebox-flash-images` file, so hardcoding `BAREBOX_IMAGES` as done here is unneeded.

meta-lxatac-bsp/recipes-core/images/emmc-boot-image/lxatac/genimage.config
```diff
@@ -18,7 +18,7 @@ image @IMAGE@fip {
     fip {
         fw-config = "stm32mp157c-lxa-tac-fw-config.dtb"
         hw-config = "stm32mp157c-lxa-tac-fw-config.dtb"
-        nt-fw = "barebox-stm32mp15xc-lxa-tac.img"
+        nt-fw = "barebox-stm32mp15xc-lxa-tac.stm32.img"
```
a3f: I think this should go away with the to-be-upstreamed OE-core recipe.

meta-lxatac-bsp/recipes-bsp/barebox/files/patches/0102-ARM-stm32mp-add-Linux-Automation-TAC-board.patch
```diff
@@ -0,0 +1,421 @@
...
+@@ -0,0 +1,38 @@
++// SPDX-License-Identifier: (GPL-2.0+ OR BSD-3-Clause)
++/*
++ * Copyright (C) 2023 Leonard Göhrs, Pengutronix
++ */
++
++#include <arm/st/stm32mp157c-lxa-tac-gen2.dts>
++#include "stm32mp15xc-lxa-tac.dtsi"
++
++/ {
```
a3f, line 220: Please cherry-pick this fixup:
```
commit 840911ef51ae1e6a8b477ea5c9e96dcbd2c9f9e2 (downstream/v2023.10.0/customers/lxa/tac)
Author:     Ahmad Fatoum <a.fatoum@pengutronix.de>
AuthorDate: Mon Mar 18 08:01:08 2024 +0100
Commit:     Ahmad Fatoum <a.fatoum@pengutronix.de>
CommitDate: Mon Mar 18 08:02:59 2024 +0100

    fixup! ARM: stm32mp: add Linux Automation TAC board
    
    There's WIP-support for the other binding in
    v2024.03.0/topic/gpio-pull-up, but until that's cleaned up and upstream,
    just add a local workaround.
    
    Reported-by: Chris Fiege <c.fiege@pengutronix.de>
    Signed-off-by: Ahmad Fatoum <a.fatoum@pengutronix.de>

diff --git a/arch/arm/dts/stm32mp15xc-lxa-tac.dtsi b/arch/arm/dts/stm32mp15xc-lxa-tac.dtsi
index 33f83e035457..00ba39a7a5be 100644
--- a/arch/arm/dts/stm32mp15xc-lxa-tac.dtsi
+++ b/arch/arm/dts/stm32mp15xc-lxa-tac.dtsi
@@ -91,6 +91,43 @@ last_chosen@10 {
 	};
 };
 
+/*
+ * There are two bindings for pin bias:
+
+ * - The pinctrl binding, which differs from controller to controller,
+ *   but is capable of representing controller specific attributes like
+ *   pull ups of varying strength. This is used via
+ *   pinctrl-names/pinctrl-[0-9]+ DT properties
+ *
+ * - Or-ing GPIO_PULL_UP or GPIO_PULL_DOWN into the GPIO flags after the
+ *   phandle, e.g.  gpios = <&gpioe 7 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+ *   Thix can expose a generic interface, but doesn't allow fine tuning.
+ *
+ * Not all Linux drivers support the second binding and barebox doesn't
+ * support it all; It requires pinctrl drivers to expose to the core
+ * a way to configure pull ups/pull downs, but we don't export this info.
+ * Instead drivers are called with the device tree node in action and
+ * instructed to apply it.
+ *
+ * We can add support for this binding for STM32 only as first step, but
+ * till then, let's just use the first binding instead to configure the
+ * GPIO pull-ups.
+ */
+&{/gpio-keys} {
+	pinctrl-names = "default";
+	pinctrl-0 = <&lxatac_gpio_key_pins>;
+};
+
+&pinctrl {
+	lxatac_gpio_key_pins: lxatac-gpio-key-pins {
+		pins {
+			pinmux = <STM32_PINMUX('E',  7, GPIO)>, /* USER_BTN2 */
+				<STM32_PINMUX('I', 11, GPIO)>; /* USER_BTN */
+			bias-pull-up;
+		};
+	};
+};
+
 &baseboard_eeprom {
 	#address-cells = <1>;
 	#size-cells = <1>;
```

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/env/nv/autoboot_abort_key
```diff
@@ -0,0 +1 @@
+ctrl-c
```
a3f, line 1: The noise on the UART should be resolved by correct the pull settings in the fixup mentioned earlier.

a3f (changes requested):

README.md
hnez: Hmm. I would not want to change this back without re-investigating the `wd -x` change again.
But I would also prefer not to do such an investigation before we have our Gen 3 compatible release.
How about a compromise in the form of a draft PR #133 that does the suggested change and has a TODO item to re-investigate `wd -x` in the README?

meta-lxatac-bsp/recipes-bsp/barebox/barebox_2023.10.0.bb
hnez: I've done so now in a separate commit. Please have a look at it.

meta-lxatac-bsp/recipes-core/images/emmc-boot-image/lxatac/genimage.config
hnez: It just spins the filename extension carousel one step further by changing the extension to `barebox-stm32mp15xc-lxa-tac.stm32`.

meta-lxatac-bsp/recipes-bsp/barebox/files/patches/0102-ARM-stm32mp-add-Linux-Automation-TAC-board.patch
hnez: done

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/env/nv/autoboot_abort_key
hnez: The UART should be unaffected, shouldn't it? It's the button pullups that have changed.
I think we could keep this change. It should make the boot more robust to all kinds of interference.

meta-lxatac-bsp/recipes-core/images/emmc-boot-image/lxatac/genimage.config
a3f: that's the correct name though. `.img` is misleading as the `.stm32` images only work by chance for your use case: The stm32 header in front is no longer parsed by TF-A and is directly executed. I think we are lucky that the header evaluates to some arithmetic instructions that don't crash.

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/env/nv/autoboot_abort_key
a3f: Yes, you're right.

meta-lxatac-bsp/recipes-bsp/barebox/barebox_2023.10.0.bb
a3f: LGTM. Thanks for giving this some wider testing!

README.md
a3f: Ok.

meta-lxatac-bsp/recipes-core/images/emmc-boot-image/lxatac/genimage.config
hnez: Huh, I don't like the sound of that. What would be the proper solution here?
Is it something we should squeeze in with this PR or address in an issue?

a3f: I can submti a follow-up PR if you test it.

hnez: deal!

a3f: I couldn't rebuild the umpf tag, so here goes my change. Please test and squash if it works:
```
commit 4d6c3ef497f7d39c15cbd3923a1c038026c2653a (HEAD)
Author:     Ahmad Fatoum <a.fatoum@pengutronix.de>
AuthorDate: Thu Apr 4 22:16:35 2024 +0200
Commit:     Ahmad Fatoum <a.fatoum@pengutronix.de>
CommitDate: Thu Apr 4 22:18:09 2024 +0200

    fixup! ARM: stm32mp: add Linux Automation TAC board
    
    ARM: stm32mp: lxa-tac: remove STM32 header from bootloader binary
    
    The STM32 header is what's parsed by the BootROM of the STM32MP1. TF-A
    used to expect the same header for the BL33 binary it loads, but it no
    longer does and instead expects a FIP image that contains amongs others
    the BL33 binary as directly executable binary without header and a
    device tree that it passes to it.
    
    For use with FIP, barebox provides a barebox-stm32mp-generic-bl33.img
    image that is not board-specific. We don't use that for the LXA TAC as
    we don't want to settle on a single device tree and instead want to
    select the correct device tree at runtime from multiple that are linked
    into barebox.
    
    Instead we continue to use the .stm32 image, which still has the STM32
    header, which is now directly executed. This didn't lead to problems so
    far, because the header happens to translate to valid ARM arithmetic
    instructions that don't break anything, but we shouldn't rely on that.
    
    Fix this by generating a plain image instead without header.
    
    Signed-off-by: Ahmad Fatoum <a.fatoum@pengutronix.de>

diff --git a/images/Makefile.stm32mp b/images/Makefile.stm32mp
index e8c7adc9d8f6..80f6b9dfc8f5 100644
--- a/images/Makefile.stm32mp
+++ b/images/Makefile.stm32mp
@@ -34,7 +34,10 @@ $(call build_stm32mp_image, CONFIG_MACH_STM32MP15XX_DKX, start_stm32mp15xx_dkx,
 $(call build_stm32mp_image, CONFIG_MACH_STM32MP15X_EV1, start_stm32mp15x_ev1, stm32mp15x-ev1)
 
 $(call build_stm32mp_image, CONFIG_MACH_LXA_MC1, start_stm32mp157c_lxa_mc1, stm32mp157c-lxa-mc1)
-$(call build_stm32mp_image, CONFIG_MACH_LXA_TAC, start_stm32mp15xc_lxa_tac, stm32mp15xc-lxa-tac)
+
+pblb-$(CONFIG_MACH_LXA_TAC) += start_stm32mp15xc_lxa_tac
+FILE_barebox-stm32mp15xc-lxa-tac.img  = start_stm32mp15xc_lxa_tac.pblb
+image-$(CONFIG_MACH_LXA_TAC) += barebox-stm32mp15xc-lxa-tac.img
 
 $(call build_stm32mp_image, CONFIG_MACH_PROTONIC_STM32MP1, start_prtt1a, prtt1a)
 $(call build_stm32mp_image, CONFIG_MACH_PROTONIC_STM32MP1, start_prtt1s, prtt1s)
```

hnez: I've applied the change and tested it now and it seems to work fine.

a3f: I guess that means this PR is ok to merge then.

a3f (approved):

PR #132, opened 2024-04-03
lxatac-core-image-base: add lrzsz for x/y/zmodem support

a3f:
Xmodem is cumbersome enough as it is without having to forward it over TCP to remote hosts, which may not always work.

Therefore, include support for the lrzsz suite of tools, so users can ssh to the lxatac and run them locally if needed.

This has been tested by copying the resulting lsz binary to the lxatac and running
```
lsz -X -vv </dev/ttyUSB2 >/dev/ttyUSB2 fw_payload.bin.out
```

### DIFF
--- a/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
+++ b/meta-lxatac-software/recipes-core/images/lxatac-core-image-base.bb
@@ -72,6 +72,7 @@ IMAGE_INSTALL:append = "\
     linux-serial-test \
     lldpd \
     lmsensors \
+    lrzsz\
     lxatac-factory-data \
     lxatac-led-setup \
     lxatac-lldpd-config \


jluebbe:
> This can be done without issues in parallel with the shell being exported and used by ser2net.

Do you mean a exported port or even with a connected client?

a3f:
> Do you mean a exported port or even with a connected client?

I had a connected client (microcom) and I had no problems using it to enable Xmodem on the DUT while above `lsz` command was waiting for the Xmodem transfer to start.

jluebbe:
How would that work? As far as I know each received byte is only read by one waiting process, in a non-deterministic way.

a3f:
Ah, right. Somehow, the latest process opening the TTY port always wins for me, so my Xmodem works fine.
I can reproduce the same with `cat /dev/ttyUSB2`: If I input characters on the ser2net console, I get the echo back only from `cat`.

I removed the claim that this works without issues with ser2net running alongside. Ok to merge now?

jluebbe (approved):

jluebbe:
Thanks for investigating the details!

Issue #128, opened 2024-03-26
The avahi daemon keeps detecting hostname conflicts


hnez:
We have noticed that `avahi-daemon` sometimes produces a lot of log output because it keeps detecting hostname conflicts:

```
Mar 20 11:39:29 lxatac-00015 avahi-daemon[306]: Host name conflict, retrying with lxatac-00015-30
Mar 20 11:39:29 lxatac-00015 avahi-daemon[306]: Registering new address record for 2a0a:edc0:1:2:1a74:e2ff:fea0:113 on tac-bridge.*.
Mar 20 11:39:29 lxatac-00015 avahi-daemon[306]: Registering new address record for 192.168.9.78 on tac-bridge.IPv4.
Mar 20 11:39:29 lxatac-00015 avahi-daemon[306]: Registering new address record for ::1 on lo.*.
Mar 20 11:39:29 lxatac-00015 avahi-daemon[306]: Registering new address record for 127.0.0.1 on lo.IPv4.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Server startup complete. Host name is lxatac-00015-30.local. Local service cookie is 1771495564.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Withdrawing address record for 192.168.9.78 on tac-bridge.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Withdrawing address record for ::1 on lo.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Withdrawing address record for 127.0.0.1 on lo.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Host name conflict, retrying with lxatac-00015-31
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Registering new address record for 2a0a:edc0:1:2:1a74:e2ff:fea0:113 on tac-bridge.*.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Registering new address record for 192.168.9.78 on tac-bridge.IPv4.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Registering new address record for ::1 on lo.*.
Mar 20 11:39:49 lxatac-00015 avahi-daemon[306]: Registering new address record for 127.0.0.1 on lo.IPv4.
[...]

PR #137, opened 2024-05-14
Update to Yocto 5.0 - Scarthgap

hnez:
This updates meta-lxatac to the yocto scarthgap release.

As part of the update we got rid of a few workarounds we needed for successful nanbield build but gained a few new workarounds needed to build for scarthgap.

The zeroconf/avahi change introduced in poky commit https://github.com/yoctoproject/poky/commit/c152a1e25c may fix the avahi daemon issues we see in #128, but I would like to test that for a bit before closing the issue.

### DIFF
--- a/meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
+++ b/meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
@@ -3,6 +3,11 @@ RRECOMMENDS:${PN}:append = "less"
 PACKAGECONFIG:append = "lz4 coredump elfutils"
 PACKAGECONFIG:remove = "networkd"
 
+# Re-enable LLMNR (but not mDNS, which is handled by avahi),
+# which is disabled by the upstream recipe when the zeroconf DISTRO_FEATURE
+# is set.
+EXTRA_OEMESON:remove = "-Ddefault-llmnr=no"
+
 do_install:append() {
     # We use system-update.target to re-partition the disk on first boot,
     # meaning we do not have all partitions available on the very first boot
--- a/meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
+++ b/meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
@@ -1,6 +1,6 @@
 RRECOMMENDS:${PN}:append = "less"
 # Enable lz4 and seccomp for systemd
-PACKAGECONFIG:append = "lz4 seccomp coredump elfutils"
+PACKAGECONFIG:append = "lz4 coredump elfutils"
 PACKAGECONFIG:remove = "networkd"
 
 do_install:append() {
--- a/meta-lxatac-software/conf/distro/tacos.conf
+++ b/meta-lxatac-software/conf/distro/tacos.conf
@@ -1,7 +1,7 @@
 DISTRO = "tacos"
 DISTRO_NAME = "TAC OS - The LXA TAC operating system"
 DISTRO_VERSION = "24.04+dev"
-DISTRO_CODENAME = "tacos-nanbield"
+DISTRO_CODENAME = "tacos-scarthgap"
 
 MAINTAINER = "Linux Automation GmbH <info@linux-automation.com>"
 
--- a/meta-lxatac-software/conf/distro/tacos.conf
+++ b/meta-lxatac-software/conf/distro/tacos.conf
@@ -7,7 +7,7 @@ MAINTAINER = "Linux Automation GmbH <info@linux-automation.com>"
 
 LOCALCONF_VERSION = "2"
 
-DISTRO_FEATURES = "argp ext2 largefile usbgadget usbhost wifi bluetooth xattr nfs zeroconf multiarch systemd usrmerge"
+DISTRO_FEATURES = "argp ext2 usbgadget usbhost wifi bluetooth xattr nfs zeroconf multiarch systemd usrmerge"
 DISTRO_FEATURES += " rauc virtualization ipv6 security seccomp alsa polkit"
 
 # Select systemd as init manager
--- a/meta-lxatac-software/recipes-connectivity/networkmanager/networkmanager_%.bbappend
+++ b/meta-lxatac-software/recipes-connectivity/networkmanager/networkmanager_%.bbappend
@@ -1,0 +1,3 @@
+# The terminal user interface is the one we document in the manual.
+# Make sure it is installed.
+PACKAGECONFIG:append = "nmtui"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "meta-ptx"]
 	path = meta-ptx
 	url = https://github.com/pengutronix/meta-ptx.git
-	branch = master
+	branch = scarthgap
 [submodule "meta-oe"]
 	path = meta-oe
 	url = https://github.com/openembedded/meta-openembedded.git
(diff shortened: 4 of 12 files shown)


meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -3,6 +3,11 @@ RRECOMMENDS:${PN}:append = "less"
 PACKAGECONFIG:append = "lz4 coredump elfutils"
 PACKAGECONFIG:remove = "networkd"
 
+# Re-enable LLMNR (but not mDNS, which is handled by avahi),
+# which is disabled by the upstream recipe when the zeroconf DISTRO_FEATURE
+# is set.
+EXTRA_OEMESON:remove = "-Ddefault-llmnr=no"
```
ejoerns, line 9: Do we actually use 'zeroconf'?

meta-lxatac-software/recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -0,0 +1,3 @@
+# The terminal user interface is the one we document in the manual.
+# Make sure it is installed.
+PACKAGECONFIG:append = "nmtui"
```
ejoerns, line 3: Seems as if the upstream change left `-Dnmtui=true` in the `EXTRA_OEMESON` . Thus potentially a patch opportunity. This change is fine however.

meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -1,7 +1,7 @@
 DISTRO = "tacos"
 DISTRO_NAME = "TAC OS - The LXA TAC operating system"
 DISTRO_VERSION = "24.04+dev"
```
ejoerns, line 3: Should this be bumped as well or only when a release is done?

meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
hnez: I use LLMNR in my setup to resolve LXA TACs in my local network (and other may as well without noticing it by using e.g. `ssh root@lxatac-00010` and having it just work), but others may use mDNS (`ssh root@lxatac-00010.local`), which means zeroconf.

`systemd-resolved` can do both (with some drawbacks on the mDNS front though, when it comes to service discovery), while avahi (which is enabled by the zeroconf `DISTRO_FEATURE`) only does mDNS but has the better D-Bus interface.

Right now we do not use any more advanced zeroconf features, like announcing services or detecting other TACs in the network, which means `systemd-resolved` would be perfectly fine for us. Because of that I initially threw out avahi support and used resolved for both.

But @jluebbe (rightfully) objected, because there are plans to do service announcements and discovery in the `tacd` in the future, which avahi currently does better.

meta-lxatac-software/conf/distro/tacos.conf
hnez: It will be bumped once we prepare the next release (and know in which month it is going to happen). Until then we carry the `+dev` suffix, which indicates that the version is built on top of release 24.04.

hnez:
I am currently tracking down another issue that needs fixing before this can be merged.

For some reason most of the `python3-twisted-*` packages in `build/tmp/work/cortexa7t2hf-neon-vfpv4-oe-linux-gnueabi/python3-twisted/24.3.0/packages-split/` end up being empty (the files all end up mushed into `packages-split/python3-twisted`), resulting in no packages being generated for e.g. `python3-twisted-conch` resulting in dependency errors later on:

```bash
$ bitbake lxatac-core-bundle-base
…
 * Solver encountered 1 problem(s):
 * Problem 1/1:
 *   - package python3-labgrid-git-r0.cortexa7t2hf-neon-vfpv4 requires python3-autobahn, but none of the providers can be installed
 *   - package python3-autobahn-23.6.2-r0.cortexa7t2hf-neon-vfpv4 requires python3-twisted, but none of the providers can be installed
 *   - conflicting requests
 *   - nothing provides python3-twisted-conch needed by python3-twisted-24.3.0-r0.cortexa7t2hf-neon-vfpv4
 *
 * Solution 1:
 *   - do not ask to install a package providing python3-labgrid
…
```

hnez:
Hi @ejoerns,

I have a suspicion on what causes the python3-twisted issues. This change in meta-oe removes a `FILES:${PN} =` assignment in the python3-twisted recipe

https://github.com/openembedded/meta-openembedded/commit/a57eea75cd93122f3dc097c55c92d8a78f19fd68#diff-6a7860eeab2c96b001eac5f36c73705cbdf45e857db2ff77e39ce8b932b5e393L78-L79

hnez:
Sorry, pressed the wrong button ("Close with comment" looks an awful lot like "discard this comment").

I think I've found the change that has cause the issue (`FILES:${PN}` no longer being overridden due to https://github.com/openembedded/meta-openembedded/commit/a57eea75cd93122f3dc097c55c92d8a78f19fd68#diff-6a7860eeab2c96b001eac5f36c73705cbdf45e857db2ff77e39ce8b932b5e393L78-L79 ) and a fix that is already in meta-oe master but not yet in the scarthgap branch https://github.com/openembedded/meta-openembedded/commit/bd1b5cde348dbc20e3d332de69617a8a95fe5ff3 .

I think I'll backport the fix like I did for the npm caching issue.

hnez:
I have added 79ccd1dfca7868d5f381470e088f54ffaa124666 now, which is not the most elegant solution, but one that at least makes the build work until the openembedded/meta-openembedded scathgap branch is updated again (which is when we can drop it again).

hnez:
I've updated meta-oe to the most recent `scarthgap` branch, which adds a single commit https://github.com/openembedded/meta-openembedded/commit/6de0ab744341ad61b0661aa28d78dc6767ce0786 (nodejs-oe-cache: fix offline install of dependencies ), which means I was able to remove the nodejs workaround I had before.
(The `python3-twisted` workaround is however still required. I guess it has lower priority).

I've also noticed that the current version of `bottom` did not build on the new rust compiler in scarthgap.
This [will be fixed](https://github.com/linux-automation/meta-lxatac/pull/139#issuecomment-2127123569) by an update in #139 but to make sure that our main branch stays buildable I've disabled `bottom` for now.

With that I think we are ready to merge this. What do you think @ejoerns / @jluebbe?

Emantor (dismissed):
The `largefile` `DISTRO_FEATURE` is the [default](https://docs.yoctoproject.org/singleindex.html#compiling-changes) in nanbield, so it can be removed.

hnez:
> The `largefile` `DISTRO_FEATURE` is the [default](https://docs.yoctoproject.org/singleindex.html#compiling-changes) in nanbield, so it can be removed.

Oh, neat. I've added a commit to remove it.

ejoerns (approved):
Fine for me.

Just to not get it lost: While checking the distro I came across another change that happened in poky (in nanbield) that we might want to adapt to (but not as part of this PR): We could also inherit the `create-spdx` class in tacos.

https://git.yoctoproject.org/poky/commit/?id=581edf20120cd383e8dea0693239629e7547bb7e

ejoerns:
Should have read this paragraph :)

> Previously, S was always created but after the recent changes it is
> not. This means the check in do_unpack_qa triggers where S is not
> created by a recipe where as it did not before. This can mean adding a
> S definition to a recipe that only uses file:// SRC_URI entries.

PR #143, opened 2024-06-12
meta-lxatac-software: labgrid-exporter: Add migration notice to config

SmithChart:
The `/etc/labgrid-exporter/environment` is migrated during rauc install. So add a note informing the user that they are free to change it.

All other labgrid exporter config files already have a notice whether they are migrated or overridden. This file was the last one without such notice.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/python/python3-labgrid/environment
+++ b/meta-lxatac-software/recipes-devtools/python/python3-labgrid/environment
@@ -1,4 +1,9 @@
+# Changes to this file will be migrated during rauc installs.
+# Feel free to change this configuration if needed.
+
 # This is a 'portable' labgrid configuration.
 # Make sure that a host 'labgrid' can be resolved using your DNS searchlist.
+# Otherwise change it to your needds.
+
 LABGRID_COORDINATOR_IP=labgrid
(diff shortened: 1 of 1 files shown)


meta-lxatac-software/recipes-devtools/python/python3-labgrid/environment
```diff
@@ -1,4 +1,8 @@
+# Changes to this file will be migrated during rauc installs.
+# Feel free to change this configuration if needed.
+
 # This is a 'portable' labgrid configuration.
 # Make sure that a host 'labgrid' can be resolved using your DNS searchlist.
+# Otherwise change it to your needds.
 LABGRID_COORDINATOR_IP=labgrid
```
hnez: I would prefer to have an empty line between the comment and the content for esthetic reasons.
But I could also live without it.

hnez (approved):

meta-lxatac-software/recipes-devtools/python/python3-labgrid/environment
SmithChart: I don't have an opinion on that detail. I've added the empty line.

PR #146, opened 2024-06-28
 ci: codespell: add spell checking CI job

hnez:
The `codespell` util allows catching typos with minimal false positives, because it uses a list of known common typos along with their correct spelling.

To limit the scope of this CI job to files we actually control, this excludes all external layers and all patch files via the `.codespellrc`.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = npm-shrinkwrap.json,*.pem,*.patch
+ignore-words-list = crate,ratatui
--- a/.github/workflows/codespell.yaml
+++ b/.github/workflows/codespell.yaml
@@ -1,0 +1,16 @@
+name: codespell
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '10 21 * * 4'
+
+jobs:
+  codespell:
+    name: codespell
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: python3 -m pip install codespell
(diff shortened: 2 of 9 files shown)


ejoerns:
Instead of providing a long exclude list, might it be easier to provide the paths to check (mainly `meta-lxatac-*` and the `README.md`) as an argument to the codespell call?

SmithChart:
> Instead of providing a long exclude list, might it be easier to provide the paths to check (mainly `meta-lxatac-*` and the `README.md`) as an argument to the codespell call?

Having an exclude list brings the advantage, that new code will be checked, even if the author does not know that it may needs to be added to a list.
The exclude list still fits on a single line on screen. I would be fine with that.

hnez:
Hi,

> > Instead of providing a long exclude list, might it be easier to provide the paths to check (mainly `meta-lxatac-*` and the `README.md`) as an argument to the codespell call?
> 
> Having an exclude list brings the advantage, that new code will be checked, even if the author does not know that it may needs to be added to a list. The exclude list still fits on a single line on screen. I would be fine with that.

looks like we've look into this PR at this at the same time. While you wrote this comment I went ahead and updated the PR to use the explicit list suggested by @ejoerns.
Do you think we should keep it that way? I don't have a strong opinion either way.

SmithChart:
> looks like we've look into this PR at this at the same time. While you wrote this comment I went ahead and updated the PR to use the explicit list suggested by @ejoerns. Do you think we should keep it that way? I don't have a strong opinion either way.

Well, bad timing. I don't have a strong opinion either.  So let's stick to the explicit list.

SmithChart (approved):

PR #153, opened 2024-07-15
 meta-lxatac-bsp: linux-lxatac: upgrade to 6.10

hnez:
umpf tag is now: 6.10/customers/lxa/lxatac/20240715-1
The tag includes the following topic branches:

  - v6.10/topic/reproducible-build
  - v6.8/customers/lxa/lxatac

This means we got rid of two topic branches, thanks to Uwe's and Oleksij's mainlining efforts.

TODO before merging:

- [x] Build and test the resulting images
- [x] Figure out what to do with the `v6.8/topic/reproducible-build` branch, which fails to merge as-is but may be updated to be mergeable instead of creating a new `v6.10` branch.


a3f (approved):
defconfig changes look ok to me.

PR #156, opened 2024-07-26
meta-lxatac-bsp: arm-trusted-firmware-a: use git instead of tarball URL

hnez:
Starting in at least late april 2024 (as indicated by [this thread][1] on the DistroKit mailing list) the old snapshot URL started responding with HTTP 401 Unauthorized errors.

Replace the tarball URL with a git URL on the same server.

[1]: https://lore.distrokit.org/distrokit/Zij_hXUXXHlJEagD@pengutronix.de/T/

### DIFF
--- a/meta-lxatac-bsp/recipes-bsp/arm-trusted-firmware-a/arm-trusted-firmware-a.inc
+++ b/meta-lxatac-bsp/recipes-bsp/arm-trusted-firmware-a/arm-trusted-firmware-a.inc
@@ -22,9 +22,9 @@ LIC_FILES_CHKSUM = "file://docs/license.rst;md5=713afe122abbe07f067f939ca3c480c5
 DEPENDS = "dtc-native"
 PROVIDES = "virtual/ARM-tf-a"
 
-SRC_URI = "https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/snapshot/${PV}.tar.gz"
+SRC_URI = "git://review.trustedfirmware.org/TF-A/trusted-firmware-a.git;protocol=https;branch=master"
 
-S = "${WORKDIR}/${PV}"
+S = "${WORKDIR}/git"
 B = "${WORKDIR}/build"
 
 inherit deploy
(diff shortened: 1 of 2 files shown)


meta-lxatac-bsp/recipes-bsp/arm-trusted-firmware-a/arm-trusted-firmware-a.inc
```diff
@@ -22,9 +22,10 @@ LIC_FILES_CHKSUM = "file://docs/license.rst;md5=713afe122abbe07f067f939ca3c480c5
 DEPENDS = "dtc-native"
 PROVIDES = "virtual/ARM-tf-a"
 
-SRC_URI = "https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/snapshot/${PV}.tar.gz"
+SRC_URI = "git://review.trustedfirmware.org/TF-A/trusted-firmware-a.git;protocol=https;branch=master"
+SRCREV = "d3e71ead6ea5bc3555ac90a446efec84ef6c6122"
```
hnez: The `SRCREV` should not be in the `.inc` file. I will fix that.

hnez:
I've assigned @Emantor now because he initially added the `arm-trusted-firmware-a.inc` (in the before-the-public-history times) and likely knows where it came from and where to check for upstream fixes.

Emantor:
This came from `meta-arm`, my suggestion would be to pull in `meta-arm` as well to use `arm-trusted-firmware-a.inc` from the layer instead of copying and modifying it.

hnez:
Oh my. I would have prepared the two-line solution proposed here, but if pulling in `meta-arm` is the cleaner solution I guess that's the way to go.
I will prepare a separate commit that pulls in `meta-arm` and uses the `trusted-firmware-a.inc` from that.

Emantor (approved):

hnez:
I've opened an issue to discuss what we want to do regarding the TF-A: #157.

Until that's decided merging this PR allows building `meta-lxatac` from scratch again, which is a good thing.

PR #159, opened 2024-07-31
CI: build: throw away the previous build artifacts

hnez:
> [!NOTE]
> I am not convinced that this is a necessary or useful change.
> I would personally prefer keeping the CI job as is.
> This PR was mainly created to check what kind of impact such a change would have on the build time.

Keeping only the sstate an downloads between runs makes the result a bit more reproducible at the cost of having to do a bit more work per-run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,8 +36,9 @@ jobs:
           # Keeping the build directory gives us incremental builds,
           # with all its benefits (speed) and drawbacks (reproducibility).
 
-          # Make sure we start with a fresh config.
-          rm -rf build/conf
+          # Make sure we start with a fresh config and and without build
+          # artifacts other than the sstate.
+          rm -rf build/conf build/tmp
 
           # oe-init-build-env implicitly cd-s into the build directory.
           cd source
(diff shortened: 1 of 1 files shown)


jluebbe:
Looks like 6:33 for https://github.com/linux-automation/meta-lxatac/actions/runs/10176907868/job/28147195616 (keep tmp/) and 8:23 for https://github.com/linux-automation/meta-lxatac/actions/runs/10177131999/job/28147911620 (remove tmp/)

hnez:
Not quite, but close. The 6:33 job actually built some things, while the 8:23 did not.

I re-ran the first job to have a fair comparison: https://github.com/linux-automation/meta-lxatac/actions/runs/10176907868

Now it's 1:46 vs. 8:23. Both are acceptable CI build times in my opinion.

I would nevertheless prefer to keep the build job as-is right now to gather some experience and to find out if it causes issues in practice. I am more than willing enough to troubleshoot those if they were to arise.

hnez:
I will close this PR until our builds fail catastrophically and I get tired of fixing them.
Then we can re-open it.

PR #162, opened 2024-08-02
README: mention the availability of a testing update channel

hnez:
And add instructions on how to switch to it / away from it.


SmithChart (approved):

hnez:
Should have mentioned that this change was inspired by @a3f, if I remember correctly.

PR #163, opened 2024-08-05
meta-lxatac-software: distro: tacos: generate and install -lic packages

hnez:
Yocto provides two different ways to deploy software license files into the operating system image.
Either using `COPY_LIC_DIRS` and `COPY_LIC_MANIFEST` or using `LICENSE_CREATE_PACKAGE`.

According to the documentation the main difference is that `LICENSE_CREATE_PACKAGE` allows adding licenses for packages installed at runtime (e.g. using `opkg`).

But this is not why this commit uses `LICENSE_CREATE_PACKAGE` instead of the other method. This commit uses `LICENSE_CREATE_PACKAGE`, because it adds a license directory per package to `/usr/share/licenses`,
containing the original license files, instead of throwing everything into `/usr/share/common-licenses`.

This PR is developed in tandem with linux-automation/tacd#76, which serves the license files via the web interface.

### DIFF
--- a/meta-lxatac-software/conf/distro/tacos.conf
+++ b/meta-lxatac-software/conf/distro/tacos.conf
@@ -46,6 +46,7 @@ INHERIT += "create-spdx"
 # Create a directory for each package in `/usr/share/licenses`
 # and place licensing information there.
 LICENSE_CREATE_PACKAGE = "1"
+COPY_LIC_MANIFEST = "1"
 
 BB_SIGNATURE_HANDLER ?= "OEEquivHash"
 BB_HASHSERVE ??= "auto"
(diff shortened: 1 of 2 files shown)


hnez:
I've talked with @jluebbe and we also want [`COPY_LIC_MANIFEST`](https://docs.yoctoproject.org/singleindex.html#term-COPY_LIC_MANIFEST) and likely a symlink that makes it available at `/usr/share/licenses`.

hnez:
I've updated this PR to also install the `license.manifest` file.
I did not go the symlinking route however, instead the list of packages is displayed in the tacd webinterface.
See linux-automation/tacd#76 for more information and a screenshot.

jluebbe (approved):

ejoerns (approved):

PR #164, opened 2024-08-06

hnez:
This adds the following new features:

  - Retry polling for update bundles with exponential backoff. This solves a problem where the first update poll after bootup is likely to fail, because the network is not yet up, and the next one only takes place hours or days later.

  - Improved error handling in the iio/dut power threads.

  - Use the modern InstallBundle and InspectBundle RAUC DBus APIs instead of the deprecated Install and Info APIs.

  - Updated rust and node dependencies.

  - A diagnostics screen that displays a lot of information on a single screen and provides a screen border that can be used to align the screen. It also allows toggling the LEDs and backlight to make sure they work correctly.

  - Notifications in the web interface that redirect the user to appropriate command lines when they try to change the active RAUC slot or enable/disable update channels in the web interface.

TODO before merging:

- [x] Update the dependencies in the bitbake files
- [x] Update the commit hash with the one from tacd/main after the [third release milestone](https://github.com/linux-automation/tacd/milestone/2) has been reached there.

### DIFF
--- a/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
+++ b/meta-lxatac-software/recipes-rust/tacd/tacd_git.bb
@@ -3,89 +3,91 @@ inherit cargo
 DEFAULT_PREFERENCE = "-1"
 
 SRC_URI += "git://github.com/linux-automation/tacd.git;protocol=https;branch=main"
-SRCREV = "e1e6ef55a8e2e9752c75e0a7debbfdb525eb1e70"
+SRCREV = "a519eb3181516ac73fa7cc7cb7aa9cccd2b2f9c0"
 S = "${WORKDIR}/git"
 CARGO_SRC_DIR = ""
 PV = "0.1.0+git${SRCPV}"
 
 SRC_URI += " \
-    crate://crates.io/addr2line/0.21.0 \
+    crate://crates.io/addr2line/0.22.0 \
     crate://crates.io/adler/1.0.2 \
     crate://crates.io/aead/0.3.2 \
     crate://crates.io/aes-gcm/0.8.0 \
     crate://crates.io/aes-soft/0.6.4 \
     crate://crates.io/aes/0.6.0 \
     crate://crates.io/aesni/0.10.0 \
-    crate://crates.io/aho-corasick/1.1.2 \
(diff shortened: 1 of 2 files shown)


hnez:
The `tacd-webinterface` build failure is interesting. The reason for it failing is that:

- [`@types/ramda`]( https://www.npmjs.com/package/@types/ramda) now depends on [`types-ramda`](https://www.npmjs.com/package/types-ramda).
- To generate a tarball name without problematic letters like "/" and "@" the former are replaced by "-" and the letter are stripped:

  https://github.com/yoctoproject/poky/blob/6ced2e7ca3b58130154035ad61685cd222909680/bitbake/lib/bb/fetch2/npm.py#L43-L51
- The `@types/ramda` and `types-ramda` tarballs end up with the same name.
- The checksum step for one of them (rightfully) fails.

hnez:
I've sent a patch for the `npm` fetcher in `bitbake` to the bitbake-devel mailing list, that should fix the `@types/ramda` vs. `types/ramda` problem.

---

The mail seems to be stuck in moderation or did not make it through some other step in the opaque process that makes appear in the web view of the mailing list. I will update this comment once it does.
~~Until then the message id is: 20240822100735.964739-1-l.goehrs@pengutronix.de.~~

---

I was not subscribed to the list, so the mail got silently dropped. I've subscribed to the list now, put on some eurobeat, got out my mullet wig and whistled some 56k modem sounds to set the mood and sent it again. This time it safely found its way through the hyperspace: https://lists.openembedded.org/g/bitbake-devel/message/16518 .

---

The fix is now in poky master and just needs to be backported to the scarthgap branch: https://github.com/yoctoproject/poky/commit/ee6bf285d7ca35ab0f76c151a36e888ad1e41c40 .

hnez:
The npm fetcher issues are worked around for now by linux-automation/tacd#78, which just downgrades the node dependencies.
This means the only thing blocking this from being mergeable is linux-automation/tacd#76.
Then we should have everything we need for a new stable release.

hnez:
Alright, linux-automation/tacd#76 is in and I have updated the commit hash here.
I've also had a quick look at the license information in the web interface and the other new features and I think we are good to go.

KarlK90 (approved):

PR #173, opened 2024-08-30
meta-lxatac-bsp: linux-lxatac: upgrade to 6.11

hnez:
umpf tag is now: 6.11/customers/lxa/lxatac/20240830-1 The tag includes the following topic branches:

  - v6.8/customers/lxa/lxatac

TODO before merging:

- [x] Test on hardware
- [x] Wait for Kernel 6.11 to be released


hnez:
I will have to investigate whether the kernel update is actually the root cause here, but the in the bundle I just tested (which includes this kernel update) the `tacd` crashed with a `SIGBUS` error in this piece of code:

https://github.com/linux-automation/tacd/blob/a25b8a72e8b7c5d10c2de39190c80b8f77d709f6/src/ui/display.rs#L99-L101

hnez:
I've just double-checked. The `SIGBUS` does not occur in Kernel 6.10. It is very unfortunate that 6.11 is already at rc5, because I will be on vacation next week and will not be able to investigate this before the release.

hnez:
I think I have found the offending commit: https://github.com/torvalds/linux/commit/86634fa4e6aeff2e190616c304156899beb4d76b the web version of the diff is not really helpful, but the command line tool shows that the merge commit contains changes which were likely introduced to fix a merge conflict.
One of these changes affects `drivers/gpu/drm/drm_fbdev_dma.c`, which (at least by name) matches the error we see.

Both parents of the merge commit compile and work fine, but the merged version does not (resulting in the previously mentioned SIGBUS error).

hnez:
Thanks to @pH5 we now know why https://github.com/torvalds/linux/commit/86634fa4e6aeff2e190616c304156899beb4d76b is the first commit to expose the problem.
It brings together https://github.com/torvalds/linux/commit/9317ff69170d and https://github.com/torvalds/linux/commit/d92a7580392a for the first time.

The first one changes `panel-mipi-dbi` from being a `drm_fbdev_generic` device to being a `drm_fbdev_dma` device.
The second one prevents `drm_fbdev_dma` devices from leaking the memory address of the framebuffer to userspace unless the `drm_kms_helper.drm_leak_fbdev_smem=1` kernel commandline option is set.

This being an issue is likely a bug in the panel-mipi-dbi driver that should be fixed there, but this workaround makes sure we can use 6.11 until the driver is fixed.

Thanks again to @pH5 for finding the two commits and the workaround.

hnez:
Hi @pH5, would you be interested reviewing for this PR?

pH5:
Looks fine to me. Maybe add a note that when the `drm_kms_helper.drm_leak_fbdev_smem=1` workaround is removed, `CONFIG_DRM_FBDEV_LEAK_PHYS_SMEM` should be disabled again as well?

hnez:
Good idea! I've added a comment  above the `BOOTSPEC_OPTIONS +=` line:

```
# Once drm_leak_fbdev_smem is removed the kernel config option
# CONFIG_DRM_FBDEV_LEAK_PHYS_SMEM should be disabled as well.
```

PR #175, opened 2024-08-30
ci: build: upload the RAUC bundle at the end of the build

hnez:
We do not want to be wasteful with the upload traffic we consume, hence why we did not upload any artifacts previously. (The sparse image for the eMMC is ~840Mb and the RAUC bundle ~290Mb).

Being able to test the the generated artifacts would however be super convenient, so upload at least the RAUC bundle for now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,6 +38,13 @@ jobs:
             emmc-boot-image \
             tf-a-stm32mp
 
+          # Make the generates images available at a less deeply nested
+          # location for the artifact upload (the directory structure is
+          # preserved in the archive).
+          # Keep in mind that we are in the `build` directory because
+          # `oe-init-build-env` implicitly `cd`s there.
+          mv tmp/deploy/images/lxatac ../images
+
       - name: Print logs of failed jobs
         if: ${{ failure() }}
         shell: python3 {0}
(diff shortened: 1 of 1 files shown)


hnez:
I've manually canceled the `build/build` job, because currently there is [a job running](
https://github.com/linux-automation/meta-lxatac/actions/runs/10627302065/job/29467587548) on the scarthgap branch right now that will update the sstate once it is done.

Waiting for that to complete and then re-starting the build process will be faster in total.

.github/workflows/build.yml
```diff
@@ -38,6 +38,11 @@ jobs:
             emmc-boot-image \
             tf-a-stm32mp
 
+          # Make the generates images available at a less deeply nested
+          # location for the artifact upload (the directory structure is
+          # preserved in the archive).
+          ln -s build/tmp/deploy/images/lxatac images
```
jluebbe: I'm not sure if symlinks work for upload-artifact. Perhaps use normal ln (hard link)?

hnez: Looks like you are correct. I've replaced the `ln -s` with a `mv` now. I could also use `cp --recursive --link` but hard linking directories is not really a thing, or is it?

jluebbe: Ah, I missed that this is a directory. `mv` sounds good.

hnez:
It took some iterations but now exactly one RAUC bundle is uploaded if the job succeeded.

jluebbe (approved):

PR #182, opened 2024-09-19
meta-lxatac-software: python3-labgrid: upgrade to version using grpc

hnez:
This marks a breaking change compared to the previous release version, as master now uses grpc instead of crossbar to communicate between the client, coordinator and exporters.

This means all of these components need to be updated to the grpc-based versions at the same time.

See https://github.com/labgrid-project/labgrid/pull/1469 for more information about this change.

Once we have tested these changes in meta-lxatac they should be moved to [meta-labgrid](https://github.com/labgrid-project/meta-labgrid/).

TODO before merging:

- [ ] Get Signed-Off-By's from @jluebbe since these changes were initially authored by him on an internal meta-lxatac customization layer.
- [x] Make sure the exporter works with other grpc based coordinators.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/python/python3-grpcio-channelz_1.62.2.bb
+++ b/meta-lxatac-software/recipes-devtools/python/python3-grpcio-channelz_1.62.2.bb
@@ -1,0 +1,16 @@
+DESCRIPTION = "Google gRPC channelz"
+HOMEPAGE = "http://www.grpc.io/"
+SECTION = "devel/python"
+
+LICENSE = "Apache-2.0"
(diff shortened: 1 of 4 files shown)


hnez:
Hi @jluebbe, how should we go forward with this?

hnez:
Hi @jluebbe, should we go forward with this PR or e.g. go via meta-labgrid instead?

jluebbe:
This should go to meta-labgrid, as we already have the `LABGRID_USE_DEVEL_VERSION` switch there that we can use.

jluebbe:
I've merged the required recipes in meta-labgrid:
https://github.com/labgrid-project/meta-labgrid/pull/58

PR #189, opened 2024-10-04
Update to Yocto 5.1 - Styhead

hnez:
This updates meta-lxatac to the yocto styhead release.

TODO before merging:

- [x] Wait for styhead to be released
- [x] Fix some of the `TODO` in the commit messages


meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -61,6 +58,8 @@ ICECC_DISABLED ??= "1"
 
 # add poky sanity bbclass
 INHERIT += "poky-sanity"
+ERROR_QA:remove = "patch-status buildpaths"
+WARN_QA:append = "patch-status buildpaths"
```
jluebbe: Perhaps limit the change to our layers only? (see https://lists.yoctoproject.org/g/yocto/message/63480)

jluebbe (changes requested):

meta-lxatac-software/conf/distro/tacos.conf
hnez: Makes sense. I've changed the respective lines.

jluebbe: Hmm, shouldn't they be in the respective layer confs? @ejoerns

ejoerns: I'd say so. Like https://web.git.yoctoproject.org/meta-arm/commit/?id=e867b0aa6a831c74cf50d23c25a74b68137bc9c3

(and with a leading space when using `append` syntax).

hnez: Sounds good. I've changed the layer configs accordingly.

hnez: The `barebox` build failed on the walnascar branch with the config moved to the layer configs, possibly because the barebox recipe itself is part of meta-oe there.
I've given up now and have just added Upstream-Stati everywhere and removed the `ERROR_QA:remove` / `WARN_QA:append`.

I don't know which recipe required the `buildpaths` exception, but I guess the CI will tell me soon.
I will update the respective recipe then.

hnez: > I don't know which recipe required the `buildpaths` exception, but I guess the CI will tell me soon.

Looks like it was not required after all. I think we are good here.

jluebbe (approved):

PR #193, opened 2024-10-18
meta-lxatac-bsp: barebox: update to 2024.10.0

hnez:
From the LXA TAC's point of view this brings two improvements:

  - Barebox now interprets `GPIO_PULL_UP` / `GPIO_PULL_DOWN` flags in gpio configurations, just like Linux does. This allows us to drop the pinctrl configuration that manually configures pull-up resistors on the two button input lines.
  - The Fastboot improvements that make `getvar` calls not interrupt the boot process anymore are now mainline and we can drop our backport.


hnez:
Hi @a3f,
could you take a look at this PR? As long as I did not throw away the wrong patches there should not be _that_ much that could have gone wrong.

a3f (approved):

PR #199, opened 2024-10-29
meta-lxatac-software: panel-mipi-dbi: cleanups for upstreaming

hnez:
We want to upstream the `panel-mipi-dbi` class and `-native` recipes (as part of internal task number 050.441).

I think [openembedded/meta-openembedded](https://github.com/openembedded/meta-openembedded) would be a good fit for such a upstreaming effort.

These are some cleanups to the recipes before making the first upstreaming attempt.

Feel free to be extra strict now, so no one is going to be mean to me on the mailing list.


ejoerns (approved):
Changes look fine to me.

Two additional note on the class:

For meta-oe `master` I guess `PANEL_FIRMWARE` needs to be looked up in `UNPACKDIR` rather than in `WORKDIR`.

I'd also suggest placing the output of `do_compile()` into `B` instead of `WORKDIR`.

hnez:
> For meta-oe `master` I guess `PANEL_FIRMWARE` needs to be looked up in `UNPACKDIR` rather than in `WORKDIR`.

That makes sense, thanks.
I've updated my styhead development branch to include this branch and the `UNPACKDIR` update:

https://github.com/linux-automation/meta-lxatac/blob/7e1b5ba2ac941f8dee0ee449dfe4e10d3a086e67/meta-lxatac-software/classes/panel-mipi-dbi.bbclass#L36-L40

> I'd also suggest placing the output of `do_compile()` into `B` instead of `WORKDIR`.

Oh, yes, thanks! Updated as well.

PR #201, opened 2024-10-30
meta-lxatac-software: bcu: prepare for upstreaming

hnez:
According to the [OpenEmbedded Layer Index](https://layers.openembedded.org/layerindex/) there is not yet an upstream [`bcu`](https://github.com/nxp-imx/bcu) recipe.
This means we should at least attempt to upstream it to e.g. `meta-openembedded`.

Do some cleanup, updating and improvements first.

Feel free to be extra strict with your review, so people do not have a reason to be mean to me on the mailing list later.


hnez:
The current version of this recipe is now in meta-openembedded master:

https://github.com/openembedded/meta-openembedded/commit/1d47796fcf9613e4ca2f2c6beeb6f447d443fca0

hnez:
Since meta-openembedded did not have any issues with this we will likely be fine.
I am going to merge it now.

PR #205, opened 2024-11-11
meta-lxatac-bsp: linux-lxatac: add aliases for can/i2c/spi and disable rtc

hnez:
The aliases ensure a stable numbering of the can / i2c / spi devices. This solves an issue we have started to observe with Linux 6.11, where the SPI bus numbers have changed/are no longer stable between boots.

The other change is disabling the RTC node, since it could not be used anyways because the RTC is missing a 32.768 kHz clock crystal.

- [x] Test on hardware and check if the SPI bus numbers match the ~~old~~ new ones in:<br/> https://github.com/linux-automation/labgrid-lxatac/pull/4


marckleinebudde (commented):
LGTM

marckleinebudde (approved):

marckleinebudde (approved):

PR #208, opened 2024-11-12
meta-lxatac-software: distro: tacos: make /srv on the rootfs read-only

hnez:
We do not want any service to write to `/srv` (in the root filesystem) before the `/srv` partition is set up and mounted.
Because if `/srv` is not empty `systemd-gpt-auto-generator` will not generate a mount unit for it.

Every now and then however some service changes and writes to `/srv/` before the partition is set up by the first boot.
Most recently podman (see https://github.com/linux-automation/meta-lxatac/pull/206 for the fix).

This is why we have added the following to the task that lead to tracking down https://github.com/linux-automation/meta-lxatac/pull/206:

> 050.600 First boot somtimes writes to /srv
> …
> [ ] lgo: Mark /srv as not-writeable in the root-fs, so that it can only be written once the /srv-partition has been mounted.

This PR implements that.

---

TODO before merging:

- [x] Make sure the permissions are actually applied

### DIFF
--- a/meta-lxatac-software/files/fs-perms-srv-ro.txt
+++ b/meta-lxatac-software/files/fs-perms-srv-ro.txt
@@ -1,0 +1,6 @@
+# This matches the original fs-perms.txt.
+# The only difference is that permissions were restricted (0755 -> 0500)
+# since the directory should only be used as a mount point.
+# This will however not stop processes running as root from writing to the
+# directory due to CAP_DAC_OVERRIDE.
+/srv				0500	root	root	false - - -
--- a/meta-lxatac-software/conf/distro/tacos.conf
+++ b/meta-lxatac-software/conf/distro/tacos.conf
@@ -40,6 +40,10 @@ USERADD_ERROR_DYNAMIC = "error"
 USERADD_UID_TABLES = "files/passwd"
 USERADD_GID_TABLES = "files/group"
 
+# Reduce read/write/directory permissions for /srv to indicate that it should
+# only be used as a mount point and that no files should be written there.
+FILESYSTEM_PERMS_TABLES += "files/fs-perms-srv-ro.txt"
+
 # Don't symlink /var/log to /var/log/volatile as we do actually want
 # persistent logging.
 VOLATILE_LOG_DIR = "no"


hnez:
This PR does indeed make `/srv` readonly

```bash
root@lxatac-00011:~# mount --mkdir /dev/mmcblk1p1 /tmp/other
root@lxatac-00011:~# cd /tmp/other/
root@lxatac-00011:/tmp/other# ls -ahl | grep srv
dr-xr-xr-x  2 root root 4.0K Nov 12 14:27 srv
```

But that does not stop software running as root from creating files there, as they have [`CAP_DAC_OVERRIDE`](https://man7.org/linux/man-pages/man7/capabilities.7.html):

> CAP_DAC_OVERRIDE
>    Bypass file read, write, and execute permission checks.
>    (DAC is an abbreviation of "discretionary access control".)

```bash
root@lxatac-00011:/tmp/other# touch srv/dac
root@lxatac-00011:/tmp/other# ls srv/dac
srv/dac
```

Having `/srv` read-only on paper is still a nice gesture though.

jluebbe:
Why not use the immutable attribute (`chattr -i /srv`) instead? You could call it from a [ROOTFS_POSTPROCESS_COMMAND](https://docs.yoctoproject.org/ref-manual/variables.html#term-ROOTFS_POSTPROCESS_COMMAND). Hopefully pseudo supports it. :)

hnez:
> Why not use the immutable attribute (`chattr -i /srv`) instead? You could call it from a [ROOTFS_POSTPROCESS_COMMAND](https://docs.yoctoproject.org/ref-manual/variables.html#term-ROOTFS_POSTPROCESS_COMMAND). Hopefully pseudo supports it. :)

Hm … I would prefer if meta-lxatac became less complex and less "special" over time. Fiddling with filesystem attributes is quite special in my opinion (I would not suspect them when something were to behave strangely on a system and look somewhere else first).

We do have a labgrid tests that triggers when the `/srv` partition is not mounted in the fully booted system, so it is not like we will not notice the issue occurring when it comes back.

meta-lxatac-software/files/fs-perms-srv-ro.txt
```diff
@@ -0,0 +1,5 @@
+# This matches the original fs-perms.txt.
+# The only difference is that write permissions for the owner were removed
+# (0755 -> 0555) to prevent anyone from writing to the rootfs /srv instead of
+# the /srv partition.
+/srv				0555	root	root	false - - -
```
jluebbe: Perhaps use `0000` to make it clearer that nobody is supposed to use this directory (except as a mount point)?

hnez: Makes sense. I've updated the PR and also removed a conflict while at it.

jluebbe (approved):

hnez:
> Perhaps use `0000` to make it clearer that nobody is supposed to use this directory (except as a mount point)?

It looks like `0000` triggers some special behavior that results in the permissions not being applied at all:

```bash
leonard@fulladder /tmp> wget "https://github-runner.pengutronix.de/artifacts/forrest-build-18HnwViPU6t38cwc/lxatac-core-bundle-base-lxatac.raucb"
leonard@fulladder /tmp> sudo mount --mkdir lxatac-core-bundle-base-lxatac.raucb /tmp/sqfs
leonard@fulladder /tmp> sudo mount --mkdir /tmp/sqfs/lxatac-core-image-base-lxatac.rootfs.ext4 /tmp/ext4
leonard@fulladder /tmp> ls -ahl /tmp/ext4/
…
drwxr-xr-x  3 root root 4,0K  9. Mär 2018  opt/
dr-xr-xr-x  2 root root 4,0K  9. Mär 2018  proc/
drwx------  2 root root 4,0K  9. Mär 2018  root/
drwxr-xr-x  2 root root 4,0K  9. Mär 2018  run/
lrwxrwxrwx  1 root root    8  9. Mär 2018  sbin -> usr/sbin/
drwxr-xr-x  2 root root 4,0K  9. Mär 2018  srv/
dr-xr-xr-x  2 root root 4,0K  9. Mär 2018  sys/
-rw-r--r--  1 root root    0  9. Mär 2018  system-update
…
```

hnez:
> It looks like 0000 triggers some special behavior that results in the permissions not being applied at all

I think I found the cause.

The `mode` field is parsed as octal number:

https://github.com/yoctoproject/poky/blob/ddfb5cd61911c58a3188c91385c7c60e6aa43ed5/meta/lib/oe/package.py#L379-L383

(Returning the integer 0 in our case)

And whenever it is used it is first checked if it is true-ish:

https://github.com/yoctoproject/poky/blob/ddfb5cd61911c58a3188c91385c7c60e6aa43ed5/meta/lib/oe/package.py#L425-L429

(Which it is not in our case)

The *intent* behind the check is to allow a mode of `-` (in that case `mode` is `None`) that keeps the permissions as-is, but duck typing makes the actual behavior more interesting once again.

hnez:
I've set the mode to `0500` now as a workaround. Since `root` does not care anyways this should not make a difference in practice.

Issue #203, opened 2024-11-05
Stable update breaks NetworkSerialPort on 24.09


a3f:
I just updated to the newest bundle on the stable channel and the exported USB console is no longer functional and ser2net emits an error message (but still starts up):
```
Invalid accepter port name/number 'telnet(rfc2217,mode=server),48431': Invalid data to parameter
```

This is because the rootfs contains ser2net 4.6.1, which requires, for our purposes, that the command line contains `,tcp`. labgrid 24.0a2.dev32 installed on the target appears to not add this argument. The commit in question is @Emantor's https://github.com/labgrid-project/labgrid/commit/a3dbfb559 first released in [v24.0](https://github.com/labgrid-project/labgrid/releases/tag/v24.0).

System info:

```
root@lxatac-00018:~# cat /etc/buildinfo 
-----------------------
Build Configuration:  |
-----------------------
DISTRO = tacos
DISTRO_VERSION = 24.09
-----------------------
Layer Revisions:      |
-----------------------
meta              = HEAD:8637aa34f0009421eb6c51a93588fe879950e1f4
meta-poky         = HEAD:8637aa34f0009421eb6c51a93588fe879950e1f4
meta-ptx          = HEAD:547b079bf309ebe1576aa5ae0d58564feb245a42
meta-oe           = HEAD:1235dd4ed4a57e67683c045ad76b6a0f9e896b45
meta-python       = HEAD:1235dd4ed4a57e67683c045ad76b6a0f9e896b45
meta-perl         = HEAD:1235dd4ed4a57e67683c045ad76b6a0f9e896b45
meta-filesystems  = HEAD:1235dd4ed4a57e67683c045ad76b6a0f9e896b45
meta-networking   = HEAD:1235dd4ed4a57e67683c045ad76b6a0f9e896b45
meta-lxatac-bsp   = HEAD:6204945f502834569d769daf071cc83606897dce
meta-lxatac-software = HEAD:6204945f502834569d769daf071cc83606897dce
meta-labgrid      = HEAD:95308c3d2ac1b101481d3f3e4590a98f1fbeb2cb
meta-rauc         = HEAD:1e3e6b334defd7fbf95cb43d23975e7b3de4b520
meta-virtualization = HEAD:54b806b1985f3989722ee308e1073530fe3328c1
meta-selinux      = HEAD:9f5a46620a07d7b15722593a33e46a4d19392b75
```

a3f:
[...]

PR #211, opened 2024-11-12
CI: add support for stable branches and minor versions

hnez:
Due to issue #203 we will need to release a v24.09.1 version of tacos.

For that we will want to create a stable-24.09 branch based on 6204945f502834569d769daf071cc83606897dce (the commit tagged as v24.09) add the required fixes and tag v24.09.1 based on it.

This is however not yet supported in our CI scripts.

Change that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,9 +4,11 @@ on:
   pull_request:
     branches:
       - scarthgap
+      - stable-*
   push:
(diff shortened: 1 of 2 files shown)


hnez:
It would be great to get this reviewed so we can start the process to push out v24.09.1 with updated labgrid.

jluebbe (approved):

PR #243, opened 2025-04-02
qa fixes and adaptions

bith3ad:
This PR is to switch to meta-arm and drop local trusted-firmware-a and fiptool recipes.


.gitmodules
```diff
@@ -26,3 +26,7 @@
 	path = meta-selinux
 	url = https://git.yoctoproject.org/meta-selinux
 	branch = scarthgap
+[submodule "meta-arm"]
+	path = meta-arm
+	url = https://git.yoctoproject.org/meta-arm
+	branch = scarthgap
```
jluebbe: Perhaps use https://github.com/jonmason/meta-arm so that dependabot can generate better changelogs?

hnez (approved):
I was most concerned with the witch to using the meta-arm provided `trusted-firmware-a`.
Mostly due to the clock setup done by the TF-A. But I have checked `/sys/kernel/debug/clk/clk_summary` before and after the change and did not see any significant changes, so I think we are good here.

The rest looks good to me as well.

bith3ad:
> I was most concerned with the witch to using the meta-arm provided `trusted-firmware-a`. Mostly due to the clock setup done by the TF-A. But I have checked `/sys/kernel/debug/clk/clk_summary` before and after the change and did not see any significant changes, so I think we are good here.
> 
> The rest looks good to me as well.

I did the same :-)

PR #234, opened 2025-02-04
meta-lxatac-bsp: linux-lxatac: upgrade to 6.14

hnez:
umpf tag is now: `6.14/customers/lxa/lxatac/20250204-1` The tag includes the following topic branches:

  - v6.11/topic/reproducible-build

See what it does not include? Any LXA TAC specific patches. Thanks to @marckleinebudde we are now mainline 🎉!

TODO before merging:

- [x] Test on hardware.
- [x] Check if `drm_kms_helper.drm_leak_fbdev_smem=1` kernel commandline option is still required and remove if it is not.
      I have checked. It is still required.
- [x] Wait for https://github.com/linux-automation/meta-lxatac/pull/228 to be merged first.
- [x] Wait for 6.14 to be released.

PR #246, opened 2025-04-22
meta-lxatac-bsp: linux: add dwc2 patch that fixes device freeze as gadget

hnez:
This integrates a patch by @mgrzeschik, which should fix an issue with the TAC running as USB Gadget (e.g. as storage gadget, emulating a USB thumb drive), where some sequences of operations may lead to the whole device freezing and subsequently being rebooted by the watchdog.

TODO before merging:

- [x] This PR is based on https://github.com/linux-automation/meta-lxatac/pull/234, which should be merged first.

@mgrzeschik: it looks like you are not in in the linux-automation organization, so I can not mark you as reviewer. Could you give this PR a look anyways?


hnez:
I've tried a sequence of `tac-gadget-storage`, `tac-gadget-remove` and USB disconnects that would have previously triggered the issue and could not reproduce it anymore with this patch applied.

SmithChart (approved):
I have run the reproducer that also does the power-off / power-on sequence on the USB host port. It can not produce the hang anymore. So, lgtm!

PR #249, opened 2025-04-25
Software fixes for 25.04 release

hnez:
These are some small fixes that came up while preparing the `styhead` and subsequent `walnascar` Yocto release upgrades.
They are however not specific to these software releases, so we can (and should) already merge them into the `scarthgap` branch.

### DIFF
--- a/meta-lxatac-software/recipes-core/lxatac-profile/lxatac-profile/01-labgrid.sh
+++ b/meta-lxatac-software/recipes-core/lxatac-profile/lxatac-profile/01-labgrid.sh
@@ -2,4 +2,4 @@
 
 # export labgrid environment to be used with labgrid-client on the lxatac
 source /etc/labgrid/environment
-export LG_CROSSBAR="ws://${LABGRID_COORDINATOR_IP:?}:${LABGRID_COORDINATOR_PORT:?}/ws"
+export LG_COORDINATOR="${LABGRID_COORDINATOR_IP:?}:${LABGRID_COORDINATOR_PORT:?}"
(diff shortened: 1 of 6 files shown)


SmithChart:
The hostname-related changes look good to me. Thx!

meta-lxatac-software/recipes-core/lxatac-profile/lxatac-profile/01-labgrid.sh
```diff
@@ -2,4 +2,4 @@
 
 # export labgrid environment to be used with labgrid-client on the lxatac
 source /etc/labgrid/environment
-export LG_CROSSBAR="ws://${LABGRID_COORDINATOR_IP:?}:${LABGRID_COORDINATOR_PORT:?}/ws"
+export LG_COORDINATOR="${LABGRID_COORDINATOR_IP:?}:${LABGRID_COORDINATOR_PORT:?}"
```
SmithChart, line 5: Please add a note to the commit message that we leave the two variables separate, so we can remember in a few years why it is this way.

SmithChart (changes requested):
One Comment. Looks good otherwise.

meta-lxatac-software/recipes-core/lxatac-profile/lxatac-profile/01-labgrid.sh
hnez: Sounds good. Done.

SmithChart (approved):

PR #250, opened 2025-04-28
Update to Yocto 5.2 - Walnascar

hnez:
This updates meta-lxatac to the yocto walnascar release.

TODO before merging:

- [x] Merge https://github.com/linux-automation/meta-lxatac/pull/189 first, which updates to the previous release (styhead).
- [x] Update the local `walnascar` branch to the merge commit of the #189 merge.
- [x] Rebase this PR on top of the updates `walnascar` branch.

TODO after merging:

- [x] Change the meta-lxatac default branch to `walnascar`.


meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -8,6 +8,8 @@ PACKAGECONFIG:remove = "networkd"
 # is set.
 EXTRA_OEMESON:remove = "-Ddefault-llmnr=no"
 
+WATCHDOG_RUNTIME_SEC = "8"
```
jluebbe: Can't this be done via the systemd-conf package? Otherwise changes here will lead to many unnecessary rebuilds. @ejoerns?

meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -16,6 +16,7 @@ DISTRO_FEATURES = "\
     minidebuginfo \
     multiarch \
     nfs \
+    pni-names \
```
jluebbe, line 19: The first sentence of the related commit message as wrong parentheses.

jluebbe (changes requested):

meta-lxatac-software/conf/distro/tacos.conf
hnez: You are right. Thanks. Fixed.

meta-lxatac-bsp/recipes-core/systemd/systemd_%.bbappend
ejoerns: IMHO, if it's a distro-like decision, I'd find it fine to put it in the Distro (or in this .bbappend, which I find less appropriate).

If that's a per-HW-decision, the original solution of using config drop-ins should be the preferred one.

hnez: The watchdog value is a hardware-based decision. I've replaced the commit that set `WATCHDOG_RUNTIME_SEC` with one that documents why we are not using it.

hnez:
Hi @jluebbe and @ejoerns, I think we are ready to merge this one.
Can I get an approval from one of you?

jluebbe (approved):

PR #254, opened 2025-05-06
meta-lxatac-software: lxatac-core-image-base: add iproute2-bridge

hnez:
This installs the `bridge` command, which we use in our testsuite to check the correct setup of the `tac-bridge` network interface.


SmithChart (approved):
lgtm. Thx!

PR #260, opened 2025-05-23
python3-labgrid: update to new version

gumulka:
Which includes a systemd service for coordinator, which we want to have installed, but not active.

Rebase after https://github.com/labgrid-project/meta-labgrid/pull/66 is approved with new meta-labgrid revision. Do not merge beforehand. Just here to not forget about it.


hnez:
These changes landed via https://github.com/linux-automation/meta-lxatac/pull/340.

PR #261, opened 2025-05-26
lxatac-core-image-base: add Snagboot and Qualcomm downloader

a3f:
There is no Labgrid support for either yet, but we need to start somewhere, so let's add them to the image for command-line usage.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/qdl/qdl_git.bb
+++ b/meta-lxatac-software/recipes-devtools/qdl/qdl_git.bb
@@ -1,0 +1,25 @@
+SUMMARY = "Qualcomm DownLoader flashing tool"
+DESCRIPTION = "Communicate with Qualcomm SoCs to upload new software or \
+dump memory"
+HOMEPAGE = "https://github.com/linux-msm/qdl.git"
+SECTION = "devel"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=da6bfde9cb5bc5120a51775381f6edf1"
+
+DEPENDS = "libxml2 libusb1"
+
+inherit pkgconfig
+
+SRCREV = "5db7794e9fdb73ed0c45384026cd8a62b5fff786"
+SRC_URI = "git://github.com/linux-msm/${BPN}.git;branch=master;protocol=https"
+
+PV = "2.1+${SRCREV}"
+
+S = "${WORKDIR}/git"
+
+do_install () {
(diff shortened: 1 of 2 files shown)


meta-lxatac-software/recipes-devtools/qdl/qdl_git.bb
```diff
@@ -0,0 +1,25 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=da6bfde9cb5bc5120a51775381f6edf1"
+
+DEPENDS = "libxml2 libusb1"
+
+inherit pkgconfig
+
+SRCREV = "5db7794e9fdb73ed0c45384026cd8a62b5fff786"
+SRC_URI = "git://github.com/linux-msm/${BPN}.git;branch=master;protocol=https"
+
+PV = "0.0+${SRCREV}"
```
hnez: Shouldn't this be:

```suggestion
PV = "2.1+${SRCREV}"
```

Because [2.1](https://github.com/linux-msm/qdl/releases/tag/v2.1) is the most recent release and we have commits on top of it?
I honestly don't know though. This could be how git versions are usually handled.

meta-lxatac-software/recipes-devtools/qdl/qdl_git.bb
```diff
@@ -0,0 +1,25 @@
...
+
+S = "${WORKDIR}/git"
+
+do_install () {
+    oe_runmake install DESTDIR=${D} prefix=${prefix}
+}
```
hnez, line 23: I don't know by heart what the default `do_install` implementation is, but I would think that it should be possible to coax it into doing rougly the same call with `DESTDIR` and `prefix` set like above.

Or in other words: can we remove this and use the default `do_install` instead?

a3f: That's how the default looks like:
```
base_do_install() {
        :
}
```

`DESTDIR=` is used in autotools.bbclass, meson.bbclass, cmake.bbclass, ... etc. which are all not applicable to qdl with its handrolled makefile.

meta-lxatac-software/recipes-devtools/qdl/qdl_git.bb
a3f: Fixed and submitted to meta-qcom as well: https://github.com/qualcomm-linux/meta-qcom/pull/930

a3f:
Patch merged into master, so now pointing at newer revision.

hnez (approved):
Looks good, thanks!

PR #262, opened 2025-05-27
ci: larger timeouts, fewer dependabot runs and cache dependabot sstate

hnez:
This does three things that should prevent a situation like we have now, where [build jobs on our main branch](https://github.com/linux-automation/meta-lxatac/actions/runs/15267147720/job/42934774427) run into timeout:

- Upload the `sstate` generated in Dependabot pull requests. This way we do not have to re-build again once the PR is merged.
- Give the jobs more time (6h -> 12h)¹.
- ~~Update the meta-* layers on a monthly basis instead of a weekly basis.~~

---

¹) I know that these numbers sound crazy, but we _do_ build quite a lot of stuff into the LXA TAC images and the build time has to account for rebuilds from scratch, possibly with other jobs running on the host at the same time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,7 @@ jobs:
   build:
     name: build
     runs-on: [self-hosted, forrest, build]
+    timeout-minutes: 720
     if: ${{ vars.HAS_BUILD_RUNNER || github.repository == 'linux-automation/meta-lxatac' }}
     steps:
       - name: Check out the repository
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Cache Data
         env:
           CACHE_KEY: ${{ secrets.YOCTO_CACHE_KEY }}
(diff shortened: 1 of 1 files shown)


hnez:
I have canceled the build job because it does not have any meaningful sstate to work with and would not upload any sstate once it is done (6h+ later).

jluebbe:
I'd prefer to keep the building schedule the same, as that let's us see issues earlier.

hnez:
> I'd prefer to keep the building schedule the same, as that let's us see issues earlier.

Okay. I have removed the commit that changed the dependabot update interval from once a week to once a month.

jluebbe (approved):

hnez:
>     * Upload the `sstate` generated in Dependabot pull requests.
>       This way we do not have to re-build again once the PR is merged.

For the record: this [does not work](https://github.com/linux-automation/meta-lxatac/actions/runs/15326436084/job/43122034593) since Dependabot workflows run in a context like an external pull request and do not have access to the ssh key required to upload sstate, even though the branches are local to the repository.

I am not complaining here, this is absolutely the behavior one would like to see. It's just me having wrong assumptions here.

PR #268, opened 2025-06-17
meta-lxatac-bsp: linux-lxatac: add RFC sysfs GPIO patches 

hnez:
This pulls in the [RFC patches][1] sent by Bartosz Golaszewski that add a new and modernized sysfs GPIO backend.

umpf tag is now: 6.15/customers/lxa/lxatac/20250617-1
The tag includes the following topic branches:

  - v6.11/topic/reproducible-build
  - v6.14/topic/dwc2_clock_gating
  - v6.15/topic/new-sysfs-gpio

This is a draft PR that will likely never be merged and only intended to have a place for testing the new sysfs GPIO backend.

[1]: https://lore.kernel.org/all/20250610-gpio-sysfs-chip-export-v1-0-a8c7aa4478b1@linaro.org/

PR #277, opened 2025-07-14
meta-lxatac-bsp: linux-lxatac: upgrade to 6.16 with new sysfs-gpio

hnez:
This is an updated version of #268, which is based on [v4 of the respective patch series](https://lore.kernel.org/all/20250704-gpio-sysfs-chip-export-v4-0-9289d8758243@linaro.org/).

This is a messy draft PR, which is mostly here to have a CI-built version of the images and to track the status of the patch series.

It includes a drive-by upgrade to kernel version 6.16-rc6, since the patch series did not apply on 6.15 (it did not apply on 6.16-rc6 either, but at least the number of extra commits required should be smaller).


hnez:
We do not intend to merge this PR (it was only meant to test the new GPIO API) and it has served its purpose.
I will prepare a clean upgrade to 6.16 (which has since been released) in the near future.

PR #280, opened 2025-07-24
meta-lxatac-bsp: barebox-tools: install rk-usb-loader

tretter:
The rk-usb-loader is already enabled in the hosttools_defconfig and targettools_defconfig and compiled by barebox-tools.

Add it to the BAREBOX_TOOLS list to actually install it.

### DIFF
--- a/meta-lxatac-bsp/recipes-bsp/barebox/barebox-tools.bbappend
+++ b/meta-lxatac-bsp/recipes-bsp/barebox/barebox-tools.bbappend
@@ -1,0 +1,1 @@
+BAREBOX_TOOLS:append = " rk-usb-loader"


meta-lxatac-bsp/recipes-bsp/barebox/barebox-tools.bbappend
```diff
@@ -0,0 +1 @@
+BAREBOX_TOOLS:append = "rk-usb-loader"
```
Bastian-Krause: You're appending to a space-separated list without adding a space.
```suggestion
BAREBOX_TOOLS:append = " rk-usb-loader"
```

Bastian-Krause (dismissed):

hnez:
Looks good to me.

@ejoerns, do we also want to change this in the `barebox-tools.bb` in oe core?

tretter:
> @ejoerns, do we also want to change this in the `barebox-tools.bb` in oe core?

I already sent a patch: https://lore.kernel.org/openembedded-core/20250724140957.4167089-1-m.tretter@pengutronix.de/

hnez (approved):

PR #284, opened 2025-08-13
meta-lxatac-bsp: linux-lxatac: 6.15 + m_can reset support

marckleinebudde:
umpf tag is now: 6.15/customers/lxa/lxatac/20250813-1 The tag includes the following topic branches:

  - v6.14/topic/dwc2_clock_gating
  - v6.15/topic/stm32mp15-m_can-reset
  - v6.15/topic/m_can-reset


meta-lxatac-bsp/recipes-kernel/linux/files/patches/series.inc
```diff
@@ -1,26 +1,47 @@
 # umpf-base: v6.15
 # umpf-flags: upstreamstatus=insert
 # umpf-name: 6.15/customers/lxa/lxatac
```
hnez, line 3: Since we are already on 6.15 right now I think the `upgrade to 6.15` part should be dropped from the commit message.
If you like you could upgrade to 6.16 instead though.

hnez (approved):
Thanks for investigating the reset situation and the other improvements.
This looks good to me, i think the commit message comment I left earlier does not warrant a re-spin, so let's merge it as is.

PR #292, opened 2025-09-09
meta-lxatac-bsp: barebox: update to 2025.08.0

hnez:
With the barebox release every patch we used to need has gone mainline and we can use barebox patch-free. (Yay!)

TODO before merging:

- [x] Test on hardware


marckleinebudde (approved):
LGTM

PR #296, opened 2025-09-12
tools: update: add a script that automatically generates new bb files

hnez:
This change is based on a similar pull request in https://github.com/labgrid-project/meta-labgrid/pull/74.
The script here is simplified compared to the one in the other PR, because we do not need as many update methods.

The PR also contains a GitHub workflow that automatically runs the script on a schedule and opens a pull request whenever updates are available.


hnez:
I have converted this PR to draft for now, not because I do not think that it could me merged as-is, but because I think that the tool may generate some discussions and I would like to fast-track the software updates first.
The software updates should land in the next tacos / meta-lxatac release, while the update tool does not have to.

This is why I have created #297, which only contains the recipe cleanups and software updates and which should be reviewed / merged first.

PR #300, opened 2025-09-30
meta-lxatac-bsp: barebox: update to 2025.09.0

hnez:
It looks like we forgot to also update barebox-tools in the past. Fix that along with the update.


meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/defconfig
```diff
@@ -119,10 +120,9 @@ CONFIG_FASTBOOT_BASE=y
 #
 # General Settings
 #
+CONFIG_NAME=""
```
a3f, line 124: fyi, You can add some descriptive string here and have it appear in the `version` output.

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/defconfig
```diff
@@ -322,6 +315,7 @@ CONFIG_CMD_REGULATOR=y
 # CONFIG_CMD_NVMEM is not set
 # CONFIG_CMD_VARINFO is not set
 CONFIG_CMD_VERSION=y
+# CONFIG_CMD_BFETCH is not set
```
a3f: Are you sure? :(

a3f (approved):

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/defconfig
hnez: Do you have a suggestion for a `CONFIG_NAME` that would make sense in our case?

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/defconfig
hnez: I've enabled `bfetch` now just to make you happy.

meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/defconfig
a3f: In-tree it's the name of the defconfig. Haven't thought about what would make sense out-of-tree. Maybe `meta-lxatac_defconfig`? Could be nice for products with multiple BSPs to easily differentiate between them as buildsystem version is usually just the layer hash.

PR #304, opened 2025-10-02
meta-lxatac-bsp: barebox: fix framebuffer console writes after update

a3f:
The device file has changed its name after some past update:
```
  barebox@Linux Automation Test Automation Controller (TAC) Gen 3:/ ls -l /dev/fbconsole0-1
  c-w-------              0 /dev/fbconsole0-1
  barebox@Linux Automation Test Automation Controller (TAC) Gen 3:/ devinfo fbconsole0
  Bus: platform
  Parent: fb0
  Parameters:
    active:  (type: string)
    font: VGA8x16 (type: enum)
    margin.bottom: 0 (type: uint32)
    margin.left: 0 (type: uint32)
    margin.right: 0 (type: uint32)
    margin.top: 0 (type: uint32)
    rotation: 0 (type: enum) (values: "0", "90", "180", "270")
```
I still need to determine when this happened, but for now restore the display output by using the new name.

### DIFF
--- a/meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/env/boot/root-a
+++ b/meta-lxatac-bsp/recipes-bsp/barebox/files/lxatac/env/boot/root-a
@@ -5,11 +5,11 @@ led led-status-red 0
 led led-status-green 252
 led led-status-blue 97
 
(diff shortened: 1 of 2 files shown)


hnez (approved):
I have confirmed that this is currently broken and that this PR does indeed fix the issue.

a3f:
I upgraded from 2024.09.0-20240919-1 to v2025.09.0, so the rename happend in-between.

PR #313, opened 2025-10-20
meta-lxatac-software: ripgrep: update to 15.0.0

hnez:
This does not build currently, because our version of rustc is too old to support the edition2024 feature.
This update can be applied once we leave the walnascar release behind.


hnez:
I have included this upgrade in #325 and will bring it forward there.

PR #315, opened 2025-10-27
meta-lxatac-bsp: barebox: update to 2025.10.0

hnez:
This adds exciting new features, like dm-verity support. We do not take advantage of them, but they are exciting nevertheless.

TODO before merging:

- [x] Test on hardware


a3f (approved):
> We do not take advantage of them, but they are exciting nevertheless.

I concur.

PR #317, opened 2025-11-06
meta-lxatac-software: lxatac-core-bundle-base: accept empty migrate.d dir

hnez:
Looks like I have shot myself in the foot again with a shell script.

---

Prior to this patch the part of the hook script that handles migration of user-specified files will fail if the `/etc/rauc/migrate.d` directory exists but does not contain any `*.conf` files.

Make sure that the glob expands to nothing instead of the string `/etc/rauc/migrate.d/*.conf` in this case by setting the `nullglob` option. This also removes the need for the ineffective check for the existance of the `/etc/rauc/migrate.d` directory.

### DIFF
--- a/meta-lxatac-software/recipes-core/bundles/files/hook.sh
+++ b/meta-lxatac-software/recipes-core/bundles/files/hook.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -exu -o pipefail
+shopt -s nullglob
 
 EXTRA_MIGRATE_LISTS_DIR="/etc/rauc/migrate.d"
 CERT_AVAILABLE_DIR="${RAUC_SLOT_MOUNT_POINT:?}/etc/rauc/certificates-available"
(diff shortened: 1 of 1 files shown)


michaelolbrich:
This will also affect other path expansions in the same script. There are other loops that will no longer fail the the given pattern matches no files.
If that is ok then the commit message should include that. Otherwise, enable and disable `nullglob` just for this one loop.

rohieb:
As an alternative, you could also test if `$(ls "${EXTRA_MIGRATE_LISTS_DIR}"/*.conf)` is empty before for-looping over it.

hnez:
> This will also affect other path expansions in the same script. There are other loops that will no longer fail the the given pattern matches no files. If that is ok then the commit message should include that.

I have checked the other loop and I think the behavior with `nullglob` matches my intent, that I had when I initially wrote the loop, better. I have added a comment about it to the commit message.

I actually find the default behavior quite surprising (not in a good way).

michaelolbrich:
> I actually find the default behavior quite surprising (not in a good way).

It depends on the use-case. The default uncovers typos if the glob is supposed to always match something.
But in the end, the choice was made a long time ago and changing it now would break a lot of scripts now.

hnez:
Hi, I would like to get this merged. @michaelolbrich should I change anything or could you give me an approval?

michaelolbrich (approved):
With the updated commit message, it looks good to me.

PR #319, opened 2025-11-09
meta-lxatac-software: labgrid-exporter: Add HTTP and TFTP provider

SmithChart:
meta-lxatac already ships with a TFTP-server (atftpd) and a HTTP-server (in tacd).
Let's export these as resources in labgrid.

### DIFF
--- a/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
+++ b/meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
@@ -103,3 +103,17 @@ lxatac-usb-ports-p{{idx}}.{{if}}:
       '@ID_PATH': '{{sysfs}}:1.{{if}}'
 {% endfor %}
 {% endfor %}
+
+
+## Internal TFTP Server
+tftp:
+  TFTPProvider:
+    internal: "/srv/tftp/"
+    external: "/"
+
+## Internal HTTP Server
+http:
+  HTTPProvider:
+    internal: "/srv/www/"
+    external: "http://{{hostname}}/srv/"
+


meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
```diff
@@ -103,3 +103,17 @@ lxatac-usb-ports-p{{idx}}.{{if}}:
       '@ID_PATH': '{{sysfs}}:1.{{if}}'
 {% endfor %}
 {% endfor %}
+
+
+## Interal TFTP Server
```
hnez:
```suggestion
## Internal TFTP Server
```

meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
```diff
@@ -103,3 +103,17 @@ lxatac-usb-ports-p{{idx}}.{{if}}:
...
+tftp:
+  TFTPProvider:
+    internal: "/srv/tftp/"
+    external: "/"
+
+## Internal HTTP Server
+http:
+  HTTPProvider:
+    internal: "/srv/www/"
+    external: "/srv/"
```
hnez: Should the `external` entry refer to an URL instead? That is how I read the [documentation](https://labgrid.readthedocs.io/en/latest/configuration.html#httpprovider).

hnez (changes requested):

meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
SmithChart: I was not aware, that one *can* use it with an URL.

In my configuration I am using it as shown here.
Then, my strategy uses it like this:
```
url = self.http.stage(self.target.env.config.get_image_path("ram_boot"))
# url: http://192.168.1.10/images/openwrt-24.10.2-ramips-mt7621-genexis_pulse-ex400-initramfs-kernel.bin
p = urllib.parse.urlparse(url)
# p: ParseResult(scheme='http', netloc='192.168.1.10', path='/images/openwrt-24.10.2-ramips-mt7621-genexis_pulse-ex400-initramfs-kernel.bin', params='', query='', fragment='')
```

If we want to add an host to the export we could use the LXA TACs hostname and then rely on DNS.

Bastian-Krause: Exactly. I think that's how it's meant to be used. The documentation could be better.

```suggestion
    external: "http://{{hostname}}/srv/"
```

See: https://labgrid.readthedocs.io/en/latest/configuration.html#templating-the-exporter-configuration

meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
```diff
@@ -103,3 +103,17 @@ lxatac-usb-ports-p{{idx}}.{{if}}:
       '@ID_PATH': '{{sysfs}}:1.{{if}}'
 {% endfor %}
 {% endfor %}
+
+
+## Internal TFTP Server
+tftp:
+  TFTPProvider:
+    internal: "/srv/tftp/"
+    external: "/"
```
Bastian-Krause, line 112: Did you test this? The TFTP server needs be configured according to https://labgrid.readthedocs.io/en/latest/configuration.html#providers.

SmithChart: Not sure what you mean: 
* Both snippets are directly from my production setup and are used to test devices. 
* My configuration looks like the one in the example - just without the sub-directory for the board.

meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
SmithChart: Ok, in this case the `HTTPProvider` example should be updated to use the `{{hostname}}`-template.

meta-lxatac-software/recipes-devtools/python/python3-labgrid/configuration.yaml
Bastian-Krause: I just wanted to make sure that the TFTP setup on the TAC was actually compatible and tested with labgrid (due to the symlink part being somewhat special, see documentation link above) before we export it as a resource.

Sounds like you did that, so :+1: from me.

SmithChart: Ok, yes. I've tested the `atftp` server and fetched some images via TFTP.

But that left me wondering if the `tacd` serving the HTTP directory handles symlinks correctly: And yes, it does.

Bastian-Krause (approved):

hnez (approved):

PR #324, opened 2025-12-04
meta-lxatac-bsp: linux-lxatac: re-add missing mcan patches

hnez:
I wrongly assumed that all mcan patches have made it into Linux 6.18, but it turns out some are still required to address issues with BERR counting and the BERR counter not being reset on interface down/up.

umpf tag is now: 6.18/customers/lxa/lxatac/20251204-1 The tag includes the following topic branches:

  - v6.18/topic/m_can-reset
  - v6.17/topic/lxa-usb-dt


hnez:
Hi, @marckleinebudde can you check if I caught the correct patches to make the [`test_can_berr_reset`](https://github.com/linux-automation/labgrid-lxatac/blob/ac7fabfb19968ba98a54c23cf86150cfa4f992f0/tests/test_interfaces_can.py#L86-L111) test happy again?

marckleinebudde (approved):
LGTM.

The `m_can` patches will go mainline with `v6.19`. The stm32mp1 DTS changes are mainline with `v6.18`.

PR #325, opened 2025-12-04
Update to Yocto 5.3 - Whinlatter

hnez:
This is a very early draft that barely compiles yet.

TODO before merging:

- [x] Clean up into usable commits.
- [x] Check every file we have to make sure it is still needed.
- [x] Test on hardware.
- [x] Update the whinlatter dummy branch to the most recent walnascar state.
- [x] Rebase this PR on top of it.


hnez:
Reminder: use `go-mod-update-modules.bbclass`.

hnez:
I think this is finally ready for review. If you want to test a build of this on your TAC run:

```
$ rauc-enable-cert devel.cert.pem
$ rauc install https://github-runner.pengutronix.de/artifacts/forrest-build-4Yu3DlIMy5rLLZvT/lxatac-core-bundle-base-lxatac.raucb
```

.gitmodules
```diff
@@ -1,32 +1,40 @@
...
+	url = https://git.openembedded.org/bitbake
+	branch = 2.14
+[submodule "openembedded-core"]
+	path = openembedded-core
+	url = https://git.openembedded.org/openembedded-core
+	branch = whinlatter
+[submodule "meta-yocto"]
+	path = meta-yocto
+	url = https://git.yoctoproject.org/meta-yocto
+	branch = whinlatter
```
jluebbe: This seems to be for meta-poky, but as far as I can see, we don't really use anything useful from there.

jluebbe (changes requested):

.gitmodules
ejoerns: I also don't think that you need `meta-yocto`. It includes `meta-poky` which just provides the `poky` example distro (and variants) and `meta-yocto-bsp` which provides the example `generic*` machines (and beagle play).

The reason it is in is probably `poky-sanity.bbclass`. I'd drop this instead from the distro since it just adds the poky-specific check for `REQUIRED_POKY_BBLAYERS_CONF_VERSION`. If that mechanism is found useful, it could also be imported (and renamed).

hnez: Thanks for catching this. I have removed `meta-yocto` now :+1:.

hnez:
Soo … are we good to go?

hnez:
> Soo … are we good to go?

Sorry for bringing this up again, but are we?

jluebbe (approved):

PR #328, opened 2026-01-05
meta-lxatac-software: gitlab-runner: update from 18.6.3 to 18.7.1

hnez:
This does not build currently, because our version of the go compiler (1.24.6)) is too old to build the new gitlab-runner release (which requires at least 1.24.11).
This update can be applied once we leave the walnascar release behind.


hnez:
I have included this upgrade in #325 and will bring it forward there.

PR #331, opened 2026-01-20
meta-lxatac-bsp: barebox: update to 2026.01.0

hnez:
This version of Barebox adds support for loading EFI binaries.
This should allow us to run e.g. Debian Installer images on the TAC from a USB flash stick.

TODO before merging:

- [ ] Decide if we want EFI loader support on the TAC
- [ ] Test on hardware


hnez:
I think we should skip this release. The next release will contain the patches we have here and should be right around the corner.

PR #334, opened 2026-02-16
tools: update: add a script that automatically updates bb files

hnez:
I have tried basically the same before in the form of #296.
But that approach did not receive much love from anyone but me, partly because it required invasive changes to the existing bitbake files.

I did however happily use the script and GitHub magic in the background to be notified about package upgrades.

Using the auto-updater for notifications but having to hand-craft PRs anyways is a bit of duplicate work and means that I can not be the one reviewing the changes. So I would prefer having the automation right here in this repo.

I have changed the implementation so that it no longer requires any changes to our existing recipes in the hopes of making it more acceptable this way.

A PR generated by this script looks something like this: https://github.com/hnez/meta-lxatac/pull/25.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,87 @@
+name: update recipes
+
+on:
+  schedule:
+    - cron: '10 21 * * 1'
+  workflow_dispatch:
+
+jobs:
+  update:
+    name: update recipes
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v5
+
+      - name: Configure git
+        run: |
+          git config user.email "info@linux-automation.com"
+          git config user.name "Recipe Updater Bot"
+
+      - name: Create a new branch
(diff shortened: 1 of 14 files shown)


hnez:
I am thinking about just merging this without much review and discussion.
I think it is useful and I do not want to deal with the (assumed) complexity of the yocto Auto Upgrade Helper (AUH).
I know that using the upstream solution is the right thing to do(TM) but I just don't want to do so.
I will be the one carrying the maintenance burden and deal with the PRs anyways.

If there are no strong objections from i.e. @jluebbe or @SmithChart in the next week or so I think I will just merge it.

ejoerns:
Without having had a deeper look, but, none of 

https://docs.yoctoproject.org/dev/dev-manual/upgrading-recipes.html#upgrading-recipes

is working for our use case?

If AUH is only good for notification, maybe `devtool upgrade` is worth a try?

hnez:
> Maybe `devtool upgrade` is worth a try?

It most likely is, but I don't want to be the one doing it because everything `devtool`  feels quite complex to me and not like something I would like to debug if it were to fail.
That's why I would prefer a KISS solution instead.

PR #338, opened 2026-02-25
Collect SBOM information and run checks after a build

hnez:
We want to keep track of vulnerabilities in our current releases. This means the nightly `testing` builds and the biannual stable releases (including point releases for them).
For that to work we have started archiving SBOM information for previous builds with #343 and will have to upload SBOM information for stable releases from now on.

This PR implements the counterpart of periodically running a CVE check based on these archived SBOMs to warn up if any new vulnerabilities pop up.


hnez:
I have finally finished triaging the open CVEs on the development branch and have marked the PR as ready for review.

tools/cve/cve_whatever.py
```diff
@@ -0,0 +1,145 @@
...
+            if has_changed or "timestamp" not in annotation:
+                has_changed = True
+                annotation["timestamp"] = now.isoformat()
+
+            assert annotation["statements"][0]["vulnerability"]["name"] == issue["id"]
+
+            # Recursively sort the dictionary by its keys before dumping it
+            # for maximum reproducibility.
+            # (Python dicts keep the insertion order of elements)
+            annotation = sort_dict(annotation)
```
jluebbe: Can't we use `dumps` with `sort_keys=True`?

meta-lxatac-bsp/conf/layer.conf
```diff
@@ -3,7 +3,9 @@ BBPATH .= ":${LAYERDIR}"
 
 # We have recipes-* directories, add to BBFILES
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
-            ${LAYERDIR}/recipes-*/*/*.bbappend"
+            ${LAYERDIR}/recipes-*/*/*.bbappend \
+            ${LAYERDIR}/backports-*/recipes-*/*/*.bb \
+            ${LAYERDIR}/backports-*/recipes-*/*/*.bbappend"
```
jluebbe: We should keep backports from newer layer versions separate from security fixes we integrate ourselves.

Perhaps `updates-*`?

jluebbe (changes requested):

tools/cve/cve_whatever.py
hnez: Uh, I did not know that exists. We can indeed.

meta-lxatac-bsp/conf/layer.conf
hnez: Sounds good.

hnez:
Should we fix the open CVEs as part of this PR or just merge it as-is and fix them after the fact?

SmithChart:
> Should we fix the open CVEs as part of this PR or just merge it as-is and fix them after the fact?

I would be fine to limit this PR to build the infrastructure and fix / mitigate the CVE-findings afterwards.

hnez:
I may have lost track here. We agreed to merge this as-is, right?

SmithChart:
> I may have lost track here. We agreed to merge this as-is, right?

Once we agreed on the technical details we agreed to merge with CVE-warnings in place and fix those afterwards in the usual manner.

I do not see any more technical discussion. So I guess we are good to go?

SmithChart (approved):
@hnez and I have agreed to fast-track this PR in the current form. 
We want to merge this in the current form to get the actual wok of tracking CVEs going.

If we are not happy with the actual implementation we change that later on.

PR #341, opened 2026-03-05
tools: update: also update the Linux Kernel

hnez:
This updates the `tools/update/update.py` script so we can also auto-update the Linux Kernel.

What makes the Kernel so special that we can not use the script as is? Its very heavy git history.
Cloning the kernel - even without BLOB objects is quite an expensive operation and a bit rude to do on a regular basis to whoever is hosting the respective mirror.
Since we are just interested in the current list of git tags, we can skip just about everything that a real clone involves and just implement the first step of the http git protocol by hand instead. The first step of a clone is getting a mapping from refs on the server to commit hashes - i.e. everything we need.

I have implemented that, but only use it for the Linux kernel. The other projects are not as heavy and it is quite nice to also have information like the release date (we get that from git commits) in the recipes - something we do not have with the lightweight backend.

As of now we just update to the most recent mainline release or stable release based on semver. For meta-lxatac stable release branches we will likely want to update the `version_pattern` so we only update to new stable releases of the same kernel.


tools/update/update.py
```diff
@@ -84,17 +84,151 @@ def branch_key(name):
...
+        for ref_line in ref_lines:
+            # For the smart protocol the lines may contain a NUL byte,
+            # followed by more information. We do not care about the extra information.
+            ref_line = ref_line.split(b"\0")[0].strip().decode()
+
+            if ref_line:
+                commit, ref = ref_line.split()
+                refs[ref] = commit
+
+        return refs
```
hnez: This can be replaced with `git ls-remote`.

hnez:
I have canceled the build since [actions/runs/23634863412](https://github.com/linux-automation/meta-lxatac/actions/runs/23634863412) is currently running, and will run for a few hours. That job will generate new sstate relevant to this job.

PR #342, opened 2026-03-06
software: distro: tacos: choose avahi as libnss-mdns provider

hnez:
There are two providers for libnss-mdns: avahi-libnss-mdns and mdns. The former is in oe-core and the latter in meta-oe. Choose the former.

---

Note: this PR used to also contain a commit that updates the Github Actions used, but that has been solved by #358.


hnez:
The GitHub Actions commit is no longer required with https://github.com/linux-automation/meta-lxatac/pull/358.
I will rebase this PR once that is merged and get rid of the commit.

jluebbe (approved):

PR #360, opened 2026-05-20
Update to Yocto 6.0 - Wrynose

hnez:
Update meta-lxatac to the most recent Yocto release **Wrynose**.


meta-lxatac-bsp/recipes-core/rauc/files/lxatac/system.conf
```diff
@@ -5,7 +5,7 @@ bundle-formats=-plain
 data-directory=/srv/rauc
 
 [keyring]
-directory=certificates-enabled
+directory=/etc/rauc/certificates-enabled
```
gumulka, line 8: Wouldn't this break, if sysconfdir is not /etc?

Not sure if this should be dynamic or sysconfdir in the recipe should be replaced with /etc.

hnez: In theory yes, but I think this is a theoretical concern. I'd prefer to cross that bridge once we get there.

meta-lxatac-software/conf/distro/tacos.conf
```diff
@@ -30,6 +31,7 @@ DISTRO_FEATURES = "\
     xattr \
     zeroconf \
 "
+DISTRO_FEATURES_OPTED_OUT += "ptest"
 
 # Select systemd as init manager
 INIT_MANAGER ?= "systemd"
```
ejoerns, line 37: Fortunately, the new default now. But could be kept here for clarity.

ejoerns (approved):

PR #369, opened 2026-06-12
software: lxa-iobus: Harden systemd-service

SmithChart:
This reduces the capabilities of the LXA IOBus server by using systemd-hardening features.

This change makes the file system read-only for the service, only giving it write access to some pseudo-devices in /dev and to it's cache-directory.

This also means that the service is not able to create it's cache directory themself. Thus it needs to be created during build-time.


meta-lxatac-software/recipes-devtools/python/python3-lxa-iobus_%.bbappend
```diff
@@ -8,4 +8,5 @@ SRC_URI += " \
 do_install:append() {
     install -D -m0644 ${UNPACKDIR}/80-can0-iobus.link ${D}${libdir}/systemd/network/80-can0-iobus.link
     install -D -m0644 ${UNPACKDIR}/lxa-iobus.service ${D}${systemd_system_unitdir}/lxa-iobus.service
+    install -d -m0644 ${D}/var/cache/lxa-iobus
```
hnez, line 11: I had to look it up, but I think this does not interact badly with systemds `tmpfiles.d`.
The relevant rules around `/var` are:

```
root@lxatac-00011:/usr/lib/tmpfiles.d# cat var.conf
…
q /var 0755 - - -
…
d /var/cache 0755 - - -
…
```

This should mean that `/var/cache` will be created if non-existent, but will not be cleaned up (which could have meant that `/var/cache/lxa-iobus` could just vanished at some point in time).

hnez (approved):

PR #379, opened 2026-06-30
software: systemd: disable osc-context feature

hnez:
The OSC 3008 standard injects control characters into the terminal output at certain context changes, i.e. when a system is booted and waits for a login.

This currently messes with our labgrid tests and we do not need it.

Disable the feature

[1]: https://uapi-group.org/specifications/specs/osc_context/


SmithChart (approved):
Thanks for hunting down this option. I imagine some of this info could also be useful for labgrid. But we can come back to this once labgrid supports (or ignores) this option.

PR #386, opened 2026-07-06
ci: rerun-cve-check: create annotation update PR when done

hnez:
This will open a Draft PR like this one https://github.com/linux-automation/meta-lxatac/pull/385 whenever a CVE check was performed.

These Draft PRs can obviously not be merged as is, because proper annotations have to be made by a human and in some cases we may also have to add backports instead of just arguing why a CVE does not affect us.

This nevertheless makes my life easier, because in the ideal case I just have to update the annotations, force-push and be done with it.

### DIFF
--- a/.github/workflows/rerun-cve-check.yml
+++ b/.github/workflows/rerun-cve-check.yml
@@ -50,3 +50,56 @@ jobs:
     uses: ./.github/workflows/sbom-cve-check.yml
     with:
       release: ${{ needs.targets.outputs.release-tag }}
+
+  pull-request:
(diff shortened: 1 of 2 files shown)


SmithChart (approved):
This looks like a good quality-of-life improvement for the ongoing CVE checking. Not only does it prepare the PR. But it also serves as a reminder that there is something to do.

I am not sure if annotations that default to ` "status": "not_affected",` are generally a good idea. But we are doing this to find out what's best. So let's just go another iteration.

PR #389, opened 2026-07-07
meta-lxatac-software: rauc: Migrate labgrid-coordiantor State

SmithChart:
Since some time we ship all the prerequisites to use the LXA TAC as a labrid-coordinator.

In case a user enables the labgrid-coordinator systemd-service migrate this activation into the new system during an update.

### DIFF
--- a/meta-lxatac-software/recipes-core/bundles/files/hook.sh
+++ b/meta-lxatac-software/recipes-core/bundles/files/hook.sh
@@ -95,6 +95,7 @@ case "$1" in
 		migrate /home/root/.bash_history
 		migrate /home/root/.ssh/authorized_keys
 		migrate /var/cache/lxa-iobus/lss-cache
+		migrate /etc/systemd/system/multi-user.target.wants/labgrid-coordinator.service
 
 		# Also allow the running system to specify additional files to
 		# migrate to the new slot via files in /etc/rauc/migrate.d.


meta-lxatac-software/recipes-core/bundles/files/hook.sh
hnez, line 46: I think this will result in a `return` for `/etc/systemd/system/multi-user.target.wants/labgrid-coordinator.service` since `systemctl enable` creates symlinks, which do not not match:

```
$ man test
…
     -f FILE
            Returns true if FILE is a regular file.
…
```

We could just make this `-e` instead.

SmithChart: I do not think that this is a problem:

```
chris@defiant:~/tmp$ touch test.file
chris@defiant:~/tmp$ ln -s test.file test.link
chris@defiant:~/tmp$ ls -lah test.*
-rw-rw-r-- 1 chris chris 0  7. Jul 15:52 test.file
lrwxrwxrwx 1 chris chris 9  7. Jul 15:52 test.link -> test.file
chris@defiant:~/tmp$ test -e test.file ; echo $?
0
chris@defiant:~/tmp$ test -e test.link ; echo $?
0
chris@defiant:~/tmp$ test -f test.file ; echo $?
0
chris@defiant:~/tmp$ test -f test.link ; echo $?
0
chris@defiant:~/tmp$ test -e test.nonexistent ; echo $?
1
chris@defiant:~/tmp$ test -f test.nonexistent ; echo $?
1
```

Both `-e` and `-f` seem to behave the same when testing a file or a symlink to a file.

hnez: Looks you are right and `test` follows symlinks by default.

hnez (approved):

PR #395, opened 2026-07-22
bsp: nm-config, factory-data: set tac-bridge MAC in dynamic nmconnection

hnez:
The previous method of setting the tac-bridge interface MAC address suffered from a race condition:

  1) NetworkManager creates the bridge.
  2) udev sets the correct MAC address from the .link file.
  3) NetworkManager starts a DHCP client with the current bridge MAC.

Usually this is the order that events happen in, but somtimes 2) and 3) are swapped, with the end result of the device receiving no IPv4 address.

Let NetworkManager set the correct MAC instead. This should remove the race condition.


SmithChart (approved):
Looks good to me! Thanks for fixing this.

Since it is a regression for the stable release: We should probably do a dot-release containing this fix.

jluebbe (changes requested):
We discussed internally that it using the `conf.d` defaults mechanism would be simpler than generating the full connection file.

jluebbe (approved):

PR #402, opened 2026-08-05
Prepare stable release 26.07.1

hnez:
This will be a point release that contains one regression fix, two security updates and smaller misc updates:

  - Fix a race condition in the MAC address assignment of the tac-bridge interface.
    The race condition would result in the tac-bridge sometimes not getting an IPv4 address, because DHCP request would be made with a temporary random MAC, while the interface was updated to use the actual MAC from the Linux Automation GmbH pool.

  - Update 7zip to get rid of https://github.com/advisories/GHSA-836v-xv29-x32v, which is a potential RCE that can be triggered via a crafted XZ file.
  - Update libssh to get rid of 11 CVEs. Not all of these are exploitable on the TAC, but the number of them and the point release happening anyways made it easy to fix them now.

  - Update the Linux kernel to its most recent point release and collect misc updates from our upstream layers.


SmithChart (approved):
Thanks for the backports and the MAC address fix. Looks good to me.

PR #403, opened 2026-08-05
CI: Let codespell run on everything and use ruff to format and lint Python code

SmithChart:
This PR does two things:

1) Fix typos in the repository and configure to run `codespell` und everything but a list of exceptions.
2) Lint and format existing Python-code using `ruff` and add a CI workflow to ensure linting and formatting.

### DIFF
--- a/tools/update/update.py
+++ b/tools/update/update.py
@@ -72,7 +72,7 @@ def semver_key(info):
 def branch_key(name):
     """Assign a priority to a branch name
 
-    A commit will likely be contained in multiple branches and we want to use
+    A commit will likely be contained in multiple branches, and we want to use
     the most descriptive one as SRCBRANCH in the recipes.
     Assign priorities to common branch names based on how descriptive they are.
     """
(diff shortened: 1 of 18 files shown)


tools/update/update.py
```diff
@@ -134,7 +133,7 @@ def commit_info(self, commit, basic):
             return {"commit_hash": self.refs().get(commit, commit)}
 
         else:
-            git_format, fields = zip(*self.LOG_FORMAT)
+            git_format, fields = zip(*self.LOG_FORMAT, strict=False)
```
hnez:
```suggestion
            git_format, fields = zip(*self.LOG_FORMAT, strict=True)
```

hnez (approved):

tools/update/update.py
hnez: I've gone ahead and applied this myself, because it was the only thing I could find.

SmithChart: Thanks for the review and for merging this change yourself! :)

PR #406, opened 2026-08-07
ci: rerun-cve-check: run cve check against last completed build job

hnez:
… Instead of the last successful build job.

As part of the build job sbom-cve-check is run. This usually fails, because there are nearly always some open CVEs.
This marks the whole build job as failed.

Take this run for example [31157259096](https://github.com/linux-automation/meta-lxatac/actions/runs/31157259096).
It determined that [29472301968](https://github.com/linux-automation/meta-lxatac/actions/runs/29472301968) was the most recent successful build - even though it is three weeks old and we have far more recent builds with failed CVE checks.

This means the last successful build job (including a successful CVE check) will usually be quite old and not what we want to re-run the CVE check against. Instead we want the most recent completed build job - even if it failed.

---

The other commit is not too relevant when running on the default branch but is relevant when running on a stable branch, where we would however like to use the most recent CVE annotations from the default branch instead of the old ones from the stable branch.

### DIFF
--- a/.github/workflows/rerun-cve-check.yml
+++ b/.github/workflows/rerun-cve-check.yml
@@ -25,7 +25,7 @@ jobs:
             --workflow build \
             --event push \
             --branch "${GITHUB_REF_NAME}" \
-            --status success \
+            --status completed \
             --json databaseId --jq ".[0].databaseId" \
           )" | tee -a "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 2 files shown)


SmithChart (approved):
We can not rely on the CVE check to complete without issues.  So running the Check against the last completed build sounds good.

In the end the important task is to regularly check if there are relevant CVEs we have to address. There is no prescribed process *how* to do that.

